PR #1, opened 2022-03-16
Tickets and orders

MasterpieceElbow:


README.md
```diff
@@ -1,5 +1,104 @@
...
+    - datetime field `created_at`, the datetime when the order was created,
+set `auto_add_now` to fill this field automatically during the creation.
+    - foreign key `user`, the user that created the order.
+   
+There should be implemented the string representation of the order:
+```python
+print(Order.objects.first())
+<Order: 2022-03-15 15:59:50.009379+00:00>
+```
+Orders should be ordered from the newest to the oldest by default.
```
danylott, line 17: Add extra breakline here

README.md
```diff
@@ -1,5 +1,104 @@
...
+    cinema_hall=cinema_hall)
+order = Order.objects.create(user_id=1)
+
+ticket = Ticket.objects.create(
+    movie_session=movie_session,
+    order=order,
+    row=19,
+    seat=20
+)
+ticket.clean()
```
danylott, line 51: Here the error should be during creation of ticket (if override `save` - this will happen)

services/movie.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: list = None,
+                 actors_ids: list = None):
+    movie = Movie.objects.create(title=movie_title,
+                                 description=movie_description)
+    if genres_ids:
+        for genre_id in genres_ids:
```
danylott, line 27: For loop here is unnecessary - use `movie.genres.set(genres_ids)`

services/movie.py
```diff
@@ -0,0 +1,31 @@
+from db.models import Movie
+
+
+def get_movies(genres_ids=None, actors_ids=None):
+    queryset = Movie.objects.all()
+
```
danylott, line 6: As I asked you in slack - here right now everything is ok, but we will ask students also to update this service - to add filter by title, if title is provided (`icontains`)

services/movie.py
```diff
@@ -0,0 +1,31 @@
...
+def get_movie_by_id(movie_id: int):
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: list = None,
+                 actors_ids: list = None):
+    movie = Movie.objects.create(title=movie_title,
```
danylott, line 24: Here we also ask students - to make this creation inside `transaction.atomic()`

services/movie_session.py
```diff
@@ -0,0 +1,42 @@
...
+
+def create_movie_session(movie_show_time,
+                         movie_id: int,
+                         cinema_hall_id: int):
+    MovieSession.objects.create(show_time=movie_show_time,
+                                movie_id=movie_id,
+                                cinema_hall_id=cinema_hall_id)
+
+
+def get_movies_sessions(session_date=None):
```
danylott, line 14: add type annotation here please

services/movie_session.py
```diff
@@ -0,0 +1,42 @@
...
+    queryset = MovieSession.objects.all()
+    if session_date:
+        session_date = datetime.datetime.strptime(
```
danylott, line 17: As I understand - you are doing this just to check, if date format is correct?

services/movie_session.py
```diff
@@ -0,0 +1,42 @@
...
+
+def get_movie_session_by_id(movie_session_id: int):
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time=None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None):
+    movie_session = MovieSession.objects.filter(id=session_id)
```
danylott, line 32: Redo this update please, as I said in comments

settings.py
```diff
@@ -15,6 +15,16 @@
...
-INSTALLED_APPS = ("db",)
+TIME_ZONE = "Europe/Kiev"
+
+USE_I18N = True
+
+USE_TZ = True
+
+INSTALLED_APPS = [
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
```
danylott, line 28: Remove these 2 lines from INSTALLED_APPS - students must add this on their own

tests/test_main.py
```diff
@@ -0,0 +1,369 @@
...
+)
+from services.movie_session import (
+    get_tickets_of_taken_seats
+)
+from services.user import create_user, get_user, update_user
+from services.order import create_order, get_orders
+
+
+@pytest.fixture()
+def database_data():
```
danylott, line 26: This is not a best practice usage of pytest fixtures - this style is more `unittest` like.
In pytest you'd better split everything in more specific fixtures, and if needed - make fixture depend on another fixutre.
So, 1 fixture for genres, another for actors, and after that movie fixture, that depends on both :)
Of course same with other ones.
And in tests - make each test depend only on those fixture you need in this test :)

You will see, how your code become much more clean and pretty

tests/test_main.py
```diff
@@ -0,0 +1,369 @@
...
+                          order_id=2,
+                          row=9,
+                          seat=6)
+
+
+def test_auth_user_models():
+    assert settings.AUTH_USER_MODEL == "db.User"
+
+
+@pytest.mark.django_db
```
danylott, line 112: Not to copy each time this decorator - you could use smth like this:
image

tests/test_main.py
```diff
@@ -0,0 +1,369 @@
...
+        cinema_hall=cheap,
+        movie=matrix
+    )
+
+    get_user_model().objects.create_user(username="user1",
+                                         password="pass1234")
+    get_user_model().objects.create_user(username="user2",
+                                         password="pass1234")
+
+    Order.objects.create(user_id=1)
```
danylott, line 86: This is a bit tricky to depend on these `id` - you'd better use here users, that you created earlier:
```python
# btw - it is better to split it to different fixture (and return these 2 users in order)
user1 = get_user_model().objects.create_user(username="user1",
                                         password="pass1234")

rder.objects.create(user=user1)
```

danylott (changes requested):

tests/test_main.py
```diff
@@ -15,293 +14,314 @@
...
-        seat=9
-    )
-    with pytest.raises(IntegrityError):
-        Ticket.objects.create(
-            order_id=1,
-            movie_session=movie_session,
-            row=9,
-            seat=9
-        )
+@pytestmark
```
danylott, line 150: It does not work without it?

danylott (approved):

services/movie.py
```diff
@@ -17,15 +17,17 @@ def get_movie_by_id(movie_id: int):
     return Movie.objects.get(id=movie_id)
 
 
+def get_movies_by_title(title: str):
```
danylott, line 20: This function should be removed, as I understand

PR #7, opened 2022-03-27
Solution

vldtsw:


db/models.py
```diff
@@ -45,3 +48,61 @@ class MovieSession(models.Model):
...
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
```
TPolina, line 59: It's better to use variable AUTH_USER_MODEL from settings here instead of just "User" (best practice)

db/models.py
```diff
@@ -45,3 +48,61 @@ class MovieSession(models.Model):
...
+
+    def __str__(self):
+        return f"{self.created_at}"
+
+    class Meta:
+        ordering = ["-created_at"]
```
TPolina, line 65: I am not sure, but probably you should put class Meta before methods, as this variant is more common and easier to read.

db/models.py
```diff
@@ -45,3 +48,61 @@ class MovieSession(models.Model):
...
+
+        return super(Ticket, self).save(force_insert,
+                                        force_update,
+                                        using,
+                                        update_fields)
+
+    class Meta:
+        constraints = [UniqueConstraint(
+            fields=["row", "seat", "movie_session"],
+            name="unique_ticket")]
```
TPolina, line 108: same about position of Meta class

services/order.py
```diff
@@ -0,0 +1,29 @@
+from typing import List
+from django.db import transaction
+from db.models import Order, Ticket, User, MovieSession
+
+
+def create_order(tickets: List[dict],
+                 username: User.username,
+                 date=None):
+    with transaction.atomic():
+        user = User.objects.get(username=username)
```
TPolina, line 10: Instead of User better use get_user_model (best practice)

services/order.py
```diff
@@ -0,0 +1,29 @@
...
+        order = Order.objects.create(user=user)
+
+        if date:
+            order.created_at = date
+            order.save()
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session=MovieSession.objects.get(
+                    id=ticket["movie_session"]),
```
TPolina, line 20: You can also connect movie_session to ticket through id (movie_session_id=ticket["movie_session"]). It's simpler, but your variant is also good

services/user.py
```diff
@@ -0,0 +1,52 @@
+from db.models import User
+
+
+def create_user(username,
+                password,
+                email=None,
+                first_name=None,
+                last_name=None):
+    user = User.objects.create(username=username)
```
TPolina, line 9: Instead of User better use get_user_model (best practice)

TPolina, line 9: In task we were asked to use create_user method to pass username and password to it at the same time

TPolina (changes requested):

db/models.py
vldtsw: Thanks, fixed

PR #9, opened 2022-03-27
Solution

phantom-terminal:


services/user.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def create_user(
+        username: str, password: str, email: str = None,
+        first_name: str = None, last_name: str = None):
+
+    user = get_user_model().objects.create_user(
+        username=username, password=password)
+
+    with transaction.atomic():
```
Nazar-kob, line 14: I think the transaction should be before the user is created, but I approve

Nazar-kob (approved):

db/models.py
```diff
@@ -40,8 +47,59 @@ def __str__(self):
...
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(settings.AUTH_USER_MODEL,
+                             on_delete=models.CASCADE)
+
+    def __str__(self):
+        return str(self.created_at)
+
```
MasterpieceElbow, line 65: Order's QuerySet should be ordered from newest to oldest by default.

MasterpieceElbow (changes requested):

db/models.py
phantom-terminal: Thanx I fixed it)

MasterpieceElbow (approved):

PR #11, opened 2022-03-31
Solution

Krimat:


services/movie.py
```diff
@@ -10,6 +11,9 @@ def get_movies(genres_ids=None, actors_ids=None):
     if actors_ids:
         queryset = queryset.filter(actors__id__in=actors_ids)
 
+    if title:
+        queryset = queryset.filter(title__contains=title)
+
```
Geksus, line 16: I believe case-insensitive search would be better

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+            Ticket.objects.create(order=order, **ticket_data)
+
+
+def get_orders(username=None):
+    queryset = Order.objects.all()
+
+    if username:
+        queryset = queryset.filter(user__username=username)
+
+    return queryset.order_by('-id')
```
Geksus, line 32: Why use '-id'? '-' would be of use while ordering by 'created_at'. I think ordering by 'username' or 'last_name' would be better

services/user.py
```diff
@@ -0,0 +1,56 @@
+from django.contrib.auth import get_user_model
+
+
+def create_user(
+        username,
+        password,
+        email=None,
+        first_name=None,
+        last_name=None
+):
```
Geksus, line 10: Annotations!

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def update_user(
+        user_id,
+        username=None,
+        password=None,
+        email=None,
+        first_name=None,
+        last_name=None
+):
```
Geksus, line 38: Annotations!

services/order.py
```diff
@@ -0,0 +1,32 @@
+from django.contrib.auth import get_user_model
+from django.db import transaction
+
+from db.models import Order, Ticket, MovieSession
+
+
+def create_order(tickets, username, date=None):
```
Geksus, line 7: Annotations!

tests/test_main.py
```diff
@@ -135,7 +135,8 @@ def test_order_str(orders_data):
 
 
 def test_order_ordering(orders_data):
-    assert list(Order.objects.all().values_list("id")) == [(3,), (2,), (1,)]
+    assert list(Order.objects.all().order_by('-id').values_list("id")) \
+           == [(3,), (2,), (1,)]
```
Geksus, line 139: Changing tests is usualy frowned upon

Geksus (changes requested):

tests/test_main.py
Krimat: The test is incorrect

services/order.py
Krimat: This ordering used to complete tests

Geksus (approved):

db/models.py
```diff
@@ -45,3 +54,69 @@ class MovieSession(models.Model):
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+    )
+
+    def __str__(self):
+        return f"{self.created_at}"
+
+    class Meta:
```
IvanRamyk, line 74: Why create an empty class `Meta`?

IvanRamyk (approved):
Good job!

PR #12, opened 2022-03-31
Solution

TPolina:


services/order.py
```diff
@@ -0,0 +1,33 @@
...
+
+def create_order(tickets, username, date=None):
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
+        order = Order.objects.create(user=user)
+
+        if date:
+            order.created_at = date
+
+        order.save()
```
Abnormaltype, line 14: You must save after setting another date in the `if date:` condition.

Abnormaltype (changes requested):

services/order.py
TPolina: thanks

Abnormaltype (approved):

IvanRamyk (approved):
Good job!

PR #15, opened 2022-04-09
'Solution'

cloudprogramm:


services/order.py
```diff
@@ -0,0 +1,36 @@
...
+        username: str,
+        date=None
+):
+    with transaction.atomic():
+        user = User.objects.get(username=username)
+        order = Order.objects.create(user=user)
+
+        if date:
+            order.created_at = date
+        order.save()
```
Abnormaltype, line 19: You must save after setting another date in the if date: condition.

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+from db.models import Ticket, User, Order, MovieSession
+
+
+def create_order(
+        tickets: List[dict],
+        username: str,
+        date=None
+):
+    with transaction.atomic():
+        user = User.objects.get(username=username)
```
Abnormaltype, line 14: Instead of User better use get_user_model (best practice)
Ex:
`user = get_user_model().objects.get(username=username)`

services/user.py
```diff
@@ -0,0 +1,57 @@
...
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(pk=user_id)
```
Abnormaltype, line 29: same problem

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+                movie = MovieSession.objects.get(
+                    pk=ticket["movie_session"]
+                )
+                ticket["movie_session"] = movie
+
+            Ticket.objects.create(order=order, **ticket)
+
+
+def get_orders(username: str = None):
+    queryset = Order.objects.all()
```
Abnormaltype, line 32: maybe better used the name "order"?

Abnormaltype (changes requested):

services/order.py
cloudprogramm: Thanks, got it!

liuda-chuzdiuk (approved):

MasterpieceElbow (approved):

PR #28, opened 2022-06-23
Solution

4ndyua:


db/models.py
```diff
@@ -17,7 +20,7 @@ def __str__(self):
 
 
 class Movie(models.Model):
-    title = models.CharField(max_length=255)
+    title = models.CharField(max_length=255, db_index=True)
```
Hectorovich, line 23: It works, but in videos Dania said that it`s better to set index using Meta class.

Hectorovich (changes requested):

Hectorovich (approved):

TPolina (approved):

PR #30, opened 2022-06-23
All Done.

Flashmobber:


db/models.py
```diff
@@ -45,3 +52,52 @@ class MovieSession(models.Model):
...
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(settings.AUTH_USER_MODEL,
+                             on_delete=models.CASCADE)
+
+    class Meta:
+        ordering = ["-id"]
```
DmytroSvirsa, line 67: Orders should be ordered from the newest to the oldest by default, so we need to sort by created_at

DmytroSvirsa (approved):

MasterpieceElbow (approved):

PR #39, opened 2022-06-23
Solution

Chardona:


db/models.py
```diff
@@ -17,7 +21,7 @@ def __str__(self):
 
 
 class Movie(models.Model):
-    title = models.CharField(max_length=255)
+    title = models.CharField(max_length=255, db_index=True)
```
Chardona, line 24: Is it correct? Should I make Meta Class with index?

Serhii-Leonenko: in my opinion it's enough )

db/models.py
```diff
@@ -45,3 +49,54 @@ class MovieSession(models.Model):
...
+            })
+
+    def save(self, force_insert=False, force_update=False, using=None,
+             update_fields=None):
+        self.full_clean()
+        return super(Ticket, self).\
+            save(force_insert, force_update, using, update_fields)
+
+    class Meta:
+        unique_together = ["row", "seat", "movie_session"]
```
Serhii-Leonenko, line 98: according to the README you should use UniqueConstraint

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):

MasterpieceElbow (approved):

PR #40, opened 2022-06-23
Solution

Kiawami:


services/user.py
```diff
@@ -0,0 +1,52 @@
...
+        user.email = email
+        user.save()
+
+    if first_name:
+        user.first_name = first_name
+        user.save()
+
+    if last_name:
+        user.last_name = last_name
+        user.save()
```
RomaTsybulia, line 52: the same here,  you no need many "save()"

PR #41, opened 2022-06-23
'Solution'

SergeyDev1996:


db/models.py
```diff
@@ -45,3 +50,65 @@ class MovieSession(models.Model):
 
     def __str__(self):
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+
+    @classmethod
+    def create_user(cls):
+        pass
```
rotsen18, line 59: its might be just 
```suggestion
        pass
```

db/models.py
```diff
@@ -45,3 +50,65 @@ class MovieSession(models.Model):
...
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name='unique_ticket'
+            )
+        ]
+
+    def __str__(self):
+        return f"{self.movie_session.movie.title} " \
+               f"{self.movie_session.show_time} " \
+               f"(row: {self.row}, seat: {self.seat})"
```
rotsen18, line 94: You have defined str method in MovieSesion, so it may be just
```suggestion
        return f"{self.movie_session} (row: {self.row}, seat: {self.seat})"
```

rotsen18 (approved):

TPolina (approved):

PR #42, opened 2022-06-23
Solution

RomaTsybulia:


services/user.py
```diff
@@ -0,0 +1,46 @@
...
+def create_user(username: str,
+                password: str,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None):
+    new_user = get_user_model().objects.create_user(
+        username=username,
+        password=password
+    )
+    user = get_user_model().objects.filter(id=new_user.id)
```
katevasilenko, line 13: I'm not sure, you need this filter

services/user.py
```diff
@@ -0,0 +1,46 @@
...
+                password: any = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None):
+    user = get_user_model().objects.filter(id=user_id)
+
+    if username:
+        user.update(username=username)
+    if password:
+        user_with_new_password = get_user_model().objects.get(id=user_id)
```
katevasilenko, line 38: And here too

katevasilenko (approved):

services/user.py
MasterpieceElbow, line 13: You don't need this, `.create_user()` returns already created user.

services/user.py
```diff
@@ -0,0 +1,46 @@
...
+                first_name: str = None,
+                last_name: str = None):
+    new_user = get_user_model().objects.create_user(
+        username=username,
+        password=password
+    )
+    user = get_user_model().objects.filter(id=new_user.id)
+
+    if email:
+        user.update(email=email)
```
MasterpieceElbow, line 16: You can assign email to a new_user and use `save()` at the end.

services/user.py
MasterpieceElbow, line 38: You don't need this if you use `get` for the user.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #45, opened 2022-06-25
Solution

vasylhnatiuk:


NOSSUDAK (approved):

db/models.py
```diff
@@ -45,3 +48,58 @@ class MovieSession(models.Model):
...
+             force_update=False,
+             using=None,
+             update_fields=None):
+        self.full_clean()
+        return super(Ticket, self).save(force_insert,
+                                        force_update, using, update_fields)
+
+    class Meta:
+        constraints = [models.UniqueConstraint(
+            fields=["row", "seat", "movie_session"], name="unique_elements")]
```
TPolina, line 98: `class Meta` should be placed after class attributes and before all methods

TPolina (approved):

PR #46, opened 2022-06-27
Solution

Evgeniy994:


db/models.py
```diff
@@ -45,3 +51,56 @@ class MovieSession(models.Model):
 
     def __str__(self):
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey("User", on_delete=models.CASCADE)
```
rotsen18, line 58: There is better to use settings.AUTH_USER_MODEL

rotsen18 (approved):

db/models.py
TPolina: I agree, you should do this

TPolina (changes requested):

MasterpieceElbow (approved):

PR #48, opened 2022-06-27
Develop

daria-niklaevna:


db/models.py
```diff
@@ -45,3 +54,62 @@ class MovieSession(models.Model):
...
+class Ticket(models.Model):
+    row = models.IntegerField()
+    seat = models.IntegerField()
+    movie_session = models.ForeignKey("MovieSession", on_delete=models.CASCADE)
+    order = models.ForeignKey("Order", on_delete=models.CASCADE)
+
+    def __str__(self):
+        return f"{self.movie_session.movie.title} " \
+               f"{self.movie_session.show_time} " \
+               f"(row: {self.row}, seat: {self.seat})"
```
rotsen18, line 86: You have defined str method in MovieSesion, so it may be just:
```suggestion
        return f"{self.movie_session} (row: {self.row}, seat: {self.seat})"
```

rotsen18 (approved):

db/models.py
```diff
@@ -45,3 +54,62 @@ class MovieSession(models.Model):
...
+        self.full_clean()
+        return super(Ticket, self).save(
+            force_insert, force_update, using, update_fields
+        )
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=["row", "seat", "movie_session"],
+                                    name="unique_constraints")
+        ]
```
TPolina, line 115: `class Meta `should be placed just after class attributess and before functions

TPolina (approved):

PR #51, opened 2022-06-29
'Solution'

Zikkolion:


db/models.py
```diff
@@ -45,3 +50,65 @@ class MovieSession(models.Model):
 
     def __str__(self):
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+
+    @classmethod
+    def create_user(cls):
+        pass
```
rotsen18, line 59: In this way it works also
```suggestion
class User(AbstractUser):
        pass
```

db/models.py
```diff
@@ -45,3 +50,65 @@ class MovieSession(models.Model):
...
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name='unique_ticket'
+            )
+        ]
+
+    def __str__(self):
+        return f"{self.movie_session.movie.title} " \
+               f"{self.movie_session.show_time} " \
+               f"(row: {self.row}, seat: {self.seat})"
```
rotsen18, line 94: You have defined str method in MovieSesion, so it may be just
```suggestion
        return f"{self.movie_session} (row: {self.row}, seat: {self.seat})
```

services/user.py
```diff
@@ -0,0 +1,42 @@
...
+
+from db.models import User
+
+
+def create_user(username,
+                password,
+                email=None,
+                first_name=None,
+                last_name=None):
+    user = User.objects.create_user(username=username, password=password,)
```
rotsen18, line 11: get_user_model() is better here instead User

services/user.py
```diff
@@ -0,0 +1,42 @@
...
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+
+
+def get_user(user_id):
+    return User.objects.get(id=user_id)
```
rotsen18, line 22: get_user_model() is better here instead User

rotsen18 (approved):

db/models.py
MasterpieceElbow, line 94: Agree with Taras, you can use already created movie session string representation.

services/user.py
MasterpieceElbow, line 11: Taras is right. When you need to use User model - use either `settings.AUTH_USER_MODEL` or `get_user_model()`.

MasterpieceElbow (changes requested):

PR #52, opened 2022-06-29
Solution

NOSSUDAK:


db/models.py
```diff
@@ -45,3 +55,65 @@ class MovieSession(models.Model):
...
+class Ticket(models.Model):
+    row = models.IntegerField()
+    seat = models.IntegerField()
+    movie_session = models.ForeignKey("MovieSession", on_delete=models.CASCADE)
+    order = models.ForeignKey("Order", on_delete=models.SET_NULL, null=True)
+
+    def __str__(self):
+        time_format = "%Y-%m-%d %H:%M:%S"
+        return f"{self.movie_session.movie.title} " \
+               f"{self.movie_session.show_time.strftime(time_format)} " \
```
rotsen18, line 83: You already defined __str__ method for movie session so its may be just f"{self.movie_session} ...

rotsen18 (approved):

db/models.py
```diff
@@ -45,3 +55,65 @@ class MovieSession(models.Model):
...
+               f"(row: {self.row}, seat: {self.seat})"
+
+    def clean(self):
+        if self.row > self.movie_session.cinema_hall.rows:
+            raise ValidationError({
+                "row": f"row number must be in available range: "
+                       f"(1, rows): (1, {self.movie_session.cinema_hall.rows})"
+            })
+        if self.seat > self.movie_session.cinema_hall.seats_in_row:
```
TPolina, line 92: same here with `seat`. Seat and row cannot have negative value

TPolina (approved):

PR #53, opened 2022-06-29
OrderTicketSolution

Ed12816:


db/models.py
```diff
@@ -45,3 +53,62 @@ class MovieSession(models.Model):
 
     def __str__(self):
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey("User", on_delete=models.CASCADE)
```
rotsen18, line 60: There is better to use settings.AUTH_USER_MODEL

db/models.py
```diff
@@ -45,3 +53,62 @@ class MovieSession(models.Model):
...
+        constraints = [
+            UniqueConstraint(fields=[
+                "row", "seat", "movie_session"
+            ], name="unique_row_and_seat")
+        ]
+
+    def __str__(self):
+        return f"{self.movie_session.movie.title} " \
+               f"{self.movie_session.show_time} " \
+               f"(row: {self.row}, seat: {self.seat})"
```
rotsen18, line 85: You have defined str method in MovieSesion, so it may be just 
```suggestion
        return f"{self.movie_session} (row: {self.row}, seat: {self.seat})"
```

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+            user=get_user_model().objects.get(username=username)
+        )
+
+        if date:
+            created_order.created_at = date
+
+        created_order.save()
+
+        for ticket in tickets:
+            row, seat, movie_session = ticket.values()
```
rotsen18, line 18: I'm not sure that it is good practise. What if the order of dictionary ticket.values() will change?

services/user.py
```diff
@@ -0,0 +1,39 @@
...
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
+
+
+def get_user(user_id: int):
+    return User.objects.get(id=user_id)
```
rotsen18, line 19: I think it is a better way
```suggestion
    return get_user_model().objects.get(id=user_id)
```

rotsen18 (approved):

db/models.py
MasterpieceElbow, line 60: Taras is right, It is better to use either `settings.AUTH_USER_MODEL` or `get_user_model()`.

services/order.py
MasterpieceElbow, line 18: Agree with Taras. Don't rely on the order of the dict, better to use keys.

MasterpieceElbow (changes requested):

TPolina (approved):

PR #56, opened 2022-07-08
Solution

Vasyl-Smutok:


services/order.py
```diff
@@ -0,0 +1,36 @@
...
+        date=None
+):
+    with transaction.atomic():
+        user_id = get_user_model().objects.get(username=username).id
+        new_order = Order.objects.create(user_id=user_id)
+
+        if date:
+            new_order.created_at = date
+
+        new_order.save()
```
Abnormaltype, line 19: you should use `save()` only if you change the `date`

services/user.py
```diff
@@ -0,0 +1,58 @@
...
+    if not last_name:
+        last_name = ""
+
+    return get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        email=email,
+        first_name=first_name,
+        last_name=last_name
+    )
```
Abnormaltype, line 26: 1) function should only create `user`, and `return` nothing
2) rewrite `if` conditions

services/user.py
```diff
@@ -0,0 +1,58 @@
...
+
+def update_user(
+        user_id: int,
+        username=None,
+        password=None,
+        email=None,
+        first_name=None,
+        last_name=None
+):
+    new_user = get_user_model().objects.get(id=user_id)
```
Abnormaltype, line 41: use already implement functions.

Abnormaltype (changes requested):

services/user.py
Vasyl-Smutok: done

Vasyl-Smutok:
done

PR #57, opened 2022-07-12
Solution

alix-mak:


services/movie_session.py
```diff
@@ -1,4 +1,5 @@
 from db.models import MovieSession
```
Abnormaltype, line 1: why not ?
```suggestion
from db.models import MovieSession, Ticket
```

services/order.py
```diff
@@ -0,0 +1,29 @@
...
+            order.save()
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+            )
+
+        return order
```
Abnormaltype, line 22: you don't need to return something

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+    if first_name:
+        new_user.first_name = first_name
+    if last_name:
+        new_user.last_name = last_name
+    if email:
+        new_user.email = email
+    new_user.save()
+
+    return new_user
```
Abnormaltype, line 23: you don't need to return something

Abnormaltype (changes requested):

services/movie_session.py
alix-mak: unfortunate inattention

PR #58, opened 2022-07-12
Solution

drivemax26:


services/order.py
```diff
@@ -0,0 +1,32 @@
...
+
+        for ticket_data in tickets:
+            Ticket.objects.create(
+                order=order,
+                row=ticket_data["row"],
+                seat=ticket_data["seat"],
+                movie_session_id=ticket_data["movie_session"]
+            )
+
+        return order
```
Abnormaltype, line 23: you don't need to return anything

services/user.py
```diff
@@ -0,0 +1,50 @@
...
+
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+    return user
```
Abnormaltype, line 24: you don't need to return anything

services/user.py
```diff
@@ -0,0 +1,50 @@
...
+
+def update_user(
+        user_id,
+        username=None,
+        password=None,
+        email=None,
+        first_name=None,
+        last_name=None):
+
+    user = get_user_model().objects.get(id=user_id)
```
Abnormaltype, line 39: use existing function

Abnormaltype (changes requested):

PR #60, opened 2022-07-23
Solution

Splend1ed:


Splend1ed:
That`s was experiments with git)

services/order.py
```diff
@@ -0,0 +1,29 @@
...
+def create_order(tickets: list[dict], username, date=None):
+    with transaction.atomic():
+        order = Order.objects.create(
+            user=User.objects.get(username=username)
+        )
+
+        if date:
+            order.created_at = date
+
+        order.save()
```
Abnormaltype, line 15: you should use `save()` only when you change the date

services/user.py
```diff
@@ -0,0 +1,40 @@
...
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = "",
+        first_name: str = "",
+        last_name: str = "",
+
+):
+    return User.objects.create(
```
Abnormaltype, line 13: would be better use this method `get_user_model()`

services/user.py
```diff
@@ -0,0 +1,40 @@
...
+        username=username,
+        password=make_password(password),
+        email=email,
+        first_name=first_name,
+        last_name=last_name
+    )
+
+
+def get_user(user_id: int):
+    return User.objects.get(id=user_id)
```
Abnormaltype, line 23: the same here

services/user.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def update_user(
+        user_id: int,
+        password: str = None,
+        **kwargs
+) -> User:
+
+    user = User.objects.get(id=user_id)
```
Abnormaltype, line 32: DRY) use the existing method

services/user.py
```diff
@@ -0,0 +1,40 @@
...
+    for key, value in kwargs.items():
+        setattr(user, key, value)
+
+    if password:
+        user.set_password(password)
+
+    user.save()
+    return user
```
Abnormaltype, line 40: this method should return nothing

Abnormaltype (changes requested):

PR #67, opened 2022-08-08
Solution

artemkazakov947:


services/user.py
```diff
@@ -0,0 +1,46 @@
...
+        password=password)
+
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
```
Abnormaltype, line 21: this function should return nothing

services/user.py
```diff
@@ -0,0 +1,46 @@
...
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(user_id,
+                username=None,
+                password=None,
+                email=None,
+                first_name=None,
+                last_name=None):
+    user = get_user_model().objects.get(id=user_id)
```
Abnormaltype, line 34: use exist function

services/user.py
```diff
@@ -0,0 +1,46 @@
...
+    if password:
+        user.set_password(password)
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
```
Abnormaltype, line 46: the same here

Abnormaltype (changes requested):

PR #82, opened 2022-09-20
'Solution'

dmitriyKostiuk:


db/migrations/0001_initial.py
```diff
@@ -1,58 +1,118 @@
-# Generated by Django 4.0.2 on 2022-06-15 12:37
```
IvanRamyk, line 1: Usually, it's a bad practice to delete migrations in the repository that were created before.
It's better just to create a new one and not modify an existing one.

IvanRamyk (approved):
Good job!

db/migrations/0001_initial.py
IvanRamyk: But ok, let it be

PR #92, opened 2022-10-08
Solution

vlad-milkov:


db/models.py
```diff
@@ -45,3 +55,69 @@ class MovieSession(models.Model):
...
+    )
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self):
+        return f"{self.created_at}"
+
+
+class UniqueConstraints:
```
IvanRamyk, line 73: extra class?

IvanRamyk (changes requested):

db/models.py
vlad-milkov: It was automatically created by IDE when I made a grammar mistake. I even haven't seen it. Fixed.

IvanRamyk (approved):
Well done!

PR #99, opened 2022-10-15
Solution

mltdll:


AriohBelskij (approved):

services/order.py
```diff
@@ -0,0 +1,45 @@
...
+        for ticket in tickets:
+            # Attempting to reduce number of DB calls a little.
+            # For some reason I wasn't able to use `bulk_create()` here. :(
+            if (session_id := ticket["movie_session"]) not in movie_sessions:
+                movie_sessions[session_id] = MovieSession.objects.get(
+                    id=session_id
+                )
+
+            Ticket.objects.create(
+                movie_session=movie_sessions[session_id],
```
Nattalli, line 32: could you make just `movie_session_id=session_id`?
without taking data from db: `MovieSession.objects.get(...`

services/user.py
```diff
@@ -0,0 +1,54 @@
...
+
+def update_user(
+    user_id: int,
+    username: str = None,
+    password: str = None,
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None,
+) -> None:
+    # Don't want to access DB when I don't need to.
```
Nattalli, line 35: cool!

Nattalli (changes requested):

services/order.py
mltdll: For some reason my code wasn't working when I tried to do it when I was implementing the original solution. Now it does. There was probably a typo that I wasn't able to find despite re-reading the code many times. :|

Nattalli: magic 🪄

Nattalli (approved):

PR #108, opened 2022-10-21
solution

AVyha:


services/order.py
```diff
@@ -0,0 +1,35 @@
...
+from django.db.models import QuerySet
+
+from db.models import Order, User, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
```
mltdll, line 11: It's a good practice to make functions such as this one return something.

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
+    return Order.objects.all().order_by("-user")
```
mltdll, line 35: I don't think you actually need `.all()` here.

db/models.py
```diff
@@ -45,3 +50,72 @@ class MovieSession(models.Model):
...
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket"
+            )
+        ]
+
+    def __str__(self) -> str:
+        return f"{self.movie_session.movie} " \
```
mltdll, line 84: Mate prefers the usage of `()` over `\` for line breaking.

mltdll (changes requested):

services/order.py
AVyha: So, for example, we can return an Order, but the task does not say anything about the return. I think the return is unnecessary here cause we can use `get_orders` to take information about order.

mltdll (approved):

services/order.py
mltdll: I agree that instructions don't say anything about it, but to use `get_orders()` means to make an additional query to the database. And it's best to avoid unnecessary database interactions when possible. :)

Nattalli (changes requested):

db/models.py
```diff
@@ -45,3 +50,72 @@ class MovieSession(models.Model):
...
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE
+    )
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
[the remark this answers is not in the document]
AVyha: i don't know which variant better) for me they the same)

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+    with transaction.atomic():
+        user = User.objects.get(username=username)
+
+        order = Order.objects.create(
+            user=user,
+            created_at=date
+        )
+        if date:
+            order.created_at = date
+            order.save()
```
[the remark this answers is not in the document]
AVyha: Yes. I forget about first `created_at` 😊

Nattalli (approved):

PR #109, opened 2022-10-21
'Solution'

davidkrivko:


Abnormaltype (changes requested):
all tests should are pass

services/user.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None) -> None:
+    user = User.objects.get(id=user_id)
```
Abnormaltype, line 34: use best practice way to get user

Abnormaltype (changes requested):

services/user.py
davidkrivko: which?

Abnormaltype: image

Abnormaltype (changes requested):
all tests should are pass

davidkrivko:
> all tests should are pass

I don't know what annotation these parameters have

db/models.py
```diff
@@ -45,3 +54,57 @@ class MovieSession(models.Model):
...
+            raise ValidationError(
+                {"seat": [f"seat number must be in available"
+                          f" range: (1, seats_in_row): (1, "
+                          f"{self.movie_session.cinema_hall.seats_in_row})"]})
+
+    def save(self, force_insert=False, force_update=False, using=None,
+             update_fields=None) -> super:
+        self.full_clean()
+        return super(Ticket, self).save(
+            force_insert, force_update, using, update_fields)
```
Abnormaltype, line 101: def save(
            self,
            force_insert: bool = False,
            force_update: bool = False,
            using: Any = None,
            update_fields: Any = None
    ) -> Any:
        self.full_clean()
        return super(Ticket, self).save(
            force_insert,
            force_update,
            using,
            update_fields
        )

Abnormaltype (changes requested):

PR #110, opened 2022-10-22
Solution

IvanStored:


services/order.py
```diff
@@ -0,0 +1,36 @@
...
+        ]
+
+
+def get_orders(username: str = None) -> QuerySet:
+    orders = Order.objects.all()
+
+    if username:
+        orders = orders.filter(user__username=username)
+
+    return orders.order_by("-id")
```
mltdll, line 36: This ordering should be implemented in the model, not in this function.

mltdll (changes requested):

mltdll (approved):

MasterpieceElbow (approved):
GJ!

PR #111, opened 2022-10-23
Solution

luxuriant777:


services/movie_session.py
```diff
@@ -42,3 +42,18 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
```
mltdll, line 47: It's better to write `list[Ticket]`.

services/order.py
```diff
@@ -0,0 +1,30 @@
...
+                order=new_order,
+                row=ticket_data["row"],
+                seat=ticket_data["seat"],
+                movie_session_id=ticket_data["movie_session"]
+            )
+
+        return new_order
+
+
+def get_orders(username: str = None) -> Order:
```
mltdll, line 27: This function returns `QuerySet`.

services/order.py
```diff
@@ -0,0 +1,30 @@
...
+    if username:
+        return Order.objects.filter(user__username=username)
+    return Order.objects.all().order_by("-id")
```
mltdll, line 30: `all()` is not necessary here.

services/user.py
```diff
@@ -0,0 +1,40 @@
+from db.models import User
+
+
+def create_user(
+        username: str, password: str, email: str = None,
+        first_name: str = None, last_name: str = None
+) -> None:
+
+    user = User.objects.create_user(username=username)
+    user.set_password(raw_password=password)
```
mltdll, line 10: It's possible to set password directly in `.create_user()`.

services/user.py
```diff
@@ -0,0 +1,40 @@
...
+        user_id: int, username: str = None, password: str = None,
+        email: str = None, first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = get_user(user_id)
+    if username:
+        user.username = username
+    if password:
+        user.set_password(raw_password=password)
+    if email:
```
mltdll, line 34: Since fields after this one can be blank, it might be better to use `is not None` for them.

mltdll (changes requested):

db/models.py
```diff
@@ -45,3 +51,72 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
```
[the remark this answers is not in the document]
luxuriant777: Thanks, changed.

db/models.py
```diff
@@ -45,3 +51,72 @@ class MovieSession(models.Model):
...
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="row-seat_movie_session"
+            )
+        ]
+
+    def __str__(self) -> str:
+        return f"{self.movie_session.movie.title} " \
```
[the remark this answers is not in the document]
luxuriant777: Thanks, changed.

services/movie_session.py
luxuriant777: Thanks, changed.

services/order.py
luxuriant777: Thanks, changed.

services/order.py
luxuriant777: Thanks, changed.

services/user.py
luxuriant777: Thanks, changed.

services/user.py
luxuriant777: Thanks, changed.

db/models.py
```diff
@@ -45,3 +51,72 @@ class MovieSession(models.Model):
...
+               f"{self.movie_session.show_time} " \
+               f"(row: {self.row}, seat: {self.seat})"
+
+    def clean(self) -> None:
```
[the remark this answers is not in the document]
luxuriant777: Can you, please, show me example? I checked documentation, but it was not clear for me.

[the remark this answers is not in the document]
mltdll: Looks like it's not necessary, since the `.clean()` method of the parent class is not doing anything.

services/order.py
```diff
@@ -24,7 +25,7 @@ def create_order(tickets: list, username: str, date: str = None) -> Order:
         return new_order
 
 
-def get_orders(username: str = None) -> Order:
+def get_orders(username: str = None) -> QuerySet:
     if username:
         return Order.objects.filter(user__username=username)
-    return Order.objects.all().order_by("-id")
+    return Order.objects.order_by("-id")
```
mltdll, line 31: By the way, `order_by()` is not necessary here; it seems that there was a bug that made tests fail on some students' machines, but it should be fixed by now. The tests should pass on GitHub without ordering here even if they fail on your computer.

mltdll (changes requested):

db/models.py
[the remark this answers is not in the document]
luxuriant777: Aa, ok

services/order.py
luxuriant777: It's ok) Corrected also

mltdll (approved):

services/movie_session.py
```diff
@@ -42,3 +42,18 @@ def update_movie_session(
...
+    for ticket in tickets:
+        row = ticket.row
+        seat = ticket.seat
+        result.append(
+            {
+                "row": row,
+                "seat": seat
+            }
+        )
+    return result
```
TPolina, line 59: you can use list comprehension here

services/order.py
```diff
@@ -0,0 +1,31 @@
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import User, Order, Ticket
+
+
+def create_order(tickets: list, username: str, date: str = None) -> Order:
+    with transaction.atomic():
+        user = User.objects.get(username=username)
```
TPolina, line 9: You should not refer to User directly

services/user.py
```diff
@@ -0,0 +1,45 @@
...
+def create_user(
+        username: str, password: str, email: str = None,
+        first_name: str = None, last_name: str = None
+) -> None:
+    get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        email=email or "",
+        first_name=first_name or "",
+        last_name=last_name or "",
```
TPolina, line 16: strange approach. and when some information was not passed it should be None not ""

services/user.py
```diff
@@ -0,0 +1,45 @@
...
+
+def update_user(
+        user_id: int, username: str = None, password: str = None,
+        email: str = None, first_name: str = None,
+        last_name: str = None
+) -> None:
+    if not any((username, password, email, first_name, last_name)):
+        return
+
+    with transaction.atomic():
```
TPolina, line 32: I am not sure we need transaction here

settings.py
```diff
@@ -24,5 +24,7 @@
 USE_TZ = False
 
 INSTALLED_APPS = [
-    "db",
+    "db", "django.contrib.auth", "django.contrib.contenttypes"
```
TPolina, line 27: each should be on the separate line

TPolina (changes requested):

services/movie_session.py
luxuriant777: Thanks, changed.

services/order.py
luxuriant777: Thanks, changed.

services/user.py
luxuriant777: Thanks, changed.

services/user.py
luxuriant777: Thanks, changed.

TPolina (approved):

PR #118, opened 2022-10-25
Solution

Aleksandro777:


services/user.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    new_user = User.objects.create_user(username=username, password=password)
```
mltdll, line 11: It's better to use `get_user_model()` instead of `User`.

services/user.py
```diff
@@ -0,0 +1,54 @@
...
+        new_user.first_name = first_name
+
+    if last_name:
+        new_user.last_name = last_name
+
+    new_user.save()
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
```
mltdll, line 26: Here as well.

services/user.py
```diff
@@ -0,0 +1,54 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None,
+) -> None:
+    user = User.objects.get(id=user_id)
```
mltdll, line 37: And here.

mltdll (changes requested):

mltdll (approved):

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+
+        if date:
+            order.created_at = date
+            order.save()
+
+        for ticket in tickets:
+            movie_session = ticket["movie_session"]
+
+            Ticket.objects.create(
+                movie_session=MovieSession.objects.get(id=movie_session),
```
Nattalli, line 24: bad solution. Read the first item in the checklist:
https://github.com/mate-academy/py-movie-moviesession-cinemahall/blob/master/checklist.md#code-efficiency

Nattalli: you shouldn't make one more query to DB. Remember: ForeignKey saves only the primary key to the object, not the full object. By default primary key is an ID. So it's enough to record ID only.

Nattalli (changes requested):

Nattalli (approved):

PR #119, opened 2022-10-25
Develop

Andriy-Makarenko:


services/order.py
```diff
@@ -0,0 +1,39 @@
...
+        user = get_user_model().objects.get(username=username)
+        order = Order.objects.create(user=user)
+
+        if date:
+            order.created_at = date
+            order.save()
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session=MovieSession.objects.get(
```
mltdll, line 24: You can use `movie_session_id` argument here to pass id directly.

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+                    id=ticket["movie_session"]
+                ),
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+            )
+
+
+def get_orders(username: str = None) -> Any:
+    queryset = Order.objects.all().order_by("-user")
```
mltdll, line 34: `.all()` is not necessary here... If `.order_by()` is. Otherwise it's the latter that should be removed.

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+                movie_session=MovieSession.objects.get(
+                    id=ticket["movie_session"]
+                ),
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+            )
+
+
+def get_orders(username: str = None) -> Any:
```
mltdll, line 33: Return value of this function is of type `QuerySet`.

mltdll (changes requested):

mltdll (approved):

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+            )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    queryset = Order.objects.order_by("-user")
```
MasterpieceElbow, line 33: You don't have to do `.order_by("-user")`. Probably, GitHub tests should pass without it.

MasterpieceElbow (changes requested):

services/order.py
Andriy-Makarenko: Without sorting in PiСharm, approximately 20% of test cases fail. "Wrong users order". Should I change it nevertheless?

MasterpieceElbow: Yes. Some students' interpreters work differently than GitHub's.

MasterpieceElbow (changes requested):

db/models.py
```diff
@@ -45,3 +49,72 @@ class MovieSession(models.Model):
...
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        to=User, on_delete=models.CASCADE, related_name="orders"
```
TPolina, line 61: you should refer to AUTH_USER_MODEL

db/models.py
```diff
@@ -45,3 +49,72 @@ class MovieSession(models.Model):
...
+    )
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+    class Meta:
+        ordering = ["-created_at"]
```
TPolina, line 68: class Meta should be placed before methods

db/models.py
```diff
@@ -45,3 +49,72 @@ class MovieSession(models.Model):
...
+            force_insert, force_update, using, update_fields
+        )
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_seat_for_session",
+            )
+        ]
```
TPolina, line 120: class Meta should be placed before methods

TPolina (changes requested):

TPolina (approved):
Well done!

PR #120, opened 2022-10-25
Solution)

Gr1zzZzzly:


db/models.py
```diff
@@ -45,3 +59,68 @@ class MovieSession(models.Model):
...
+    class Meta:
+        ordering = ["-created_at"]
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        to=MovieSession,
+        on_delete=models.CASCADE
+    )
+    order = models.ForeignKey(to=Order, on_delete=models.CASCADE)
```
mltdll, line 80: It might be better to specify `related_name` for foreign key fields.

db/models.py
```diff
@@ -45,3 +59,68 @@ class MovieSession(models.Model):
...
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat"],
+                name="row_seat"
+            )
+        ]
+
+    def __str__(self) -> str:
+        return f"{self.movie_session.movie.title} " \
```
mltdll, line 93: Using `\` for this is discouraged in Mate.

db/models.py
```diff
@@ -45,3 +59,68 @@ class MovieSession(models.Model):
...
+                "seat": f"seat number must be in available range: "
+                        f"(1, seats_in_row):"
+                        f" (1, {self.movie_session.cinema_hall.seats_in_row})"
+            })
+
+    def save(
+            self, force_insert: bool = False,
+            force_update: bool = False,
+            using: Any = None,
+            update_fields: Any = None) -> Any:
```
mltdll, line 118: If there's a newline after the opening parenthesis, there should also be a newline before the closing one.

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+from django.db.models import DateTimeField, QuerySet
+
+
+@transaction.atomic
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: DateTimeField = None
+) -> None:
+    user = User.objects.get(username=username)
```
mltdll, line 13: Use `get_user_model()` instead of `User`. Check out the [checklist](https://github.com/mate-academy/py-movie-moviesession-cinemahall/blob/master/checklist.md#code-efficiency).

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+    order = Order.objects.create(user=user)
+
+    if date:
+        order.created_at = date
+        order.save()
+
+    Ticket.objects.bulk_create([
+        Ticket(
+            movie_session=MovieSession.objects.get(id=ticket["movie_session"]),
```
mltdll, line 22: Use `movie_session_id` argument and pass the id directly.

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+        ) for ticket in tickets
+    ])
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+
+        return Order.objects.filter(user__username=username)
+
+    return Order.objects.all().order_by("-user__username")
```
mltdll, line 35: Tests on GitHub should pass without the `order_by()`.

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+        username=username,
+        password=password,
+        email=email,
+        first_name=first_name,
+        last_name=last_name
+    )
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(pk=user_id)
```
mltdll, line 23: Use `get_user_model()` everywhere.

mltdll (changes requested):

services/order.py
Gr1zzZzzly: ya this thing doesnt work which bulk_create, i dont know why, but if I change bulk_create to just loop all ok

mltdll: But if you use `.get()` every time, this solution won't gain anything from `bulk_create()`. It will access the database for every ticket even before it executes.

Gr1zzZzzly: ya i know , i changed it already , but my electricity is gone XD

mltdll: I hope you saved your files before it went out!

db/models.py
Gr1zzZzzly: maybe but i dont think it necessary)

services/order.py
Gr1zzZzzly: yep but it doesnt work for some reason (

mltdll: There was a bug in tests that made tests fail on some students' computers. It should be fixed by now, but it might fail on you computer if you have an older version of the repo.
Either way, it doesn't happen on GitHub.

Gr1zzZzzly: got it

mltdll (approved):

db/models.py
```diff
@@ -25,6 +36,9 @@ class Movie(models.Model):
     def __str__(self) -> str:
         return self.title
 
+    class Meta:
+        indexes = [models.Index(fields=["title"])]
```
TPolina, line 40: class Meta should be placed before methods. Fix this everywhere

db/models.py
```diff
@@ -45,3 +59,70 @@ class MovieSession(models.Model):
...
+        to=MovieSession,
+        on_delete=models.CASCADE
+    )
+    order = models.ForeignKey(to=Order, on_delete=models.CASCADE)
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(fields=["row", "seat"], name="row_seat")
```
TPolina, line 89: fields row, seat, movie_session should be unique together

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = "",
+        first_name: str = "",
+        last_name: str = ""
```
TPolina, line 11: when some information was not passed it should be None not ""

TPolina (changes requested):

services/user.py
Gr1zzZzzly: Ok i will fix this but if i use none like default variable, then I must check all attribute for existence like
if email: ...
if first_name:...
etc.
if I use just "" I don't need check it all )

TPolina: okay, I considered this solution again and think that it's also good

TPolina (approved):
Well done!

PR #121, opened 2022-10-26
Solution

shurushku:


services/user.py
```diff
@@ -0,0 +1,51 @@
...
+        first_name: str = None,
+        last_name: str = None,
+) -> None:
+    user = get_user(user_id)
+
+    if username:
+        user.username = username
+    if password:
+        user.set_password(password)
+    if isinstance(email, str):
```
mltdll, line 44: And here as well.

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None,
+) -> None:
+    get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        email=email if isinstance(email, str) else "",
+        first_name=first_name if isinstance(first_name, str) else "",
+        last_name=last_name if isinstance(first_name, str) else "",
```
mltdll, line 18: I guess `is not None` would be enough here.

mltdll (approved):

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
+                order=new_order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+            )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    queryset = Order.objects.all().order_by("-id")
```
MasterpieceElbow, line 33: You don't have to do `.order_by("-id")`. Probably, GitHub tests should pass without it.

MasterpieceElbow (changes requested):

services/order.py
shurushku: Pytest don't pass without that:
https://imgur.com/a/Q0WoaIn
Maybe I'm doing something else wrong

PR #122, opened 2022-10-26
Solution

Vasyl-Poremchuk:


services/order.py
```diff
@@ -0,0 +1,35 @@
...
+            )
+
+        return order
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
+
+    return Order.objects.order_by("-user")
```
fsocie7y, line 35: i think u shouldn't do this) this was not in the condition

fsocie7y: just return queryset

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+        tickets: list[dict[str, int]], username: str, date: str = None
+) -> Order:
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
+        order = Order.objects.create(user=user)
+
+        if date:
+            order.created_at = date
+
+        order.save()
```
fsocie7y, line 18: i think it's "order.save()" should be inside "if" statment

fsocie7y (changes requested):

fsocie7y (approved):
GGWP)

Y-Havryliv (approved):
Well done!

PR #123, opened 2022-10-27
Solution

kateryna3k:


db/models.py
```diff
@@ -17,14 +20,19 @@ def __str__(self) -> str:
...
+    title = models.CharField(max_length=255, db_index=True)
     description = models.TextField()
     actors = models.ManyToManyField(to=Actor)
     genres = models.ManyToManyField(to=Genre)
 
     def __str__(self) -> str:
         return self.title
 
+    class Meta:
+        indexes = [
```
mltdll, line 32: You're setting indexing for this field twice.

mltdll (changes requested):

mltdll (approved):

mltdll:
I think it would also be good to set `related_name` to all the foreign key and many-to-many fields, but it's hadrly required.

MasterpieceElbow (approved):
GJ!

PR #124, opened 2022-10-27
Solution

Mowka7:


db/models.py
```diff
@@ -45,3 +50,63 @@ class MovieSession(models.Model):
...
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(to=User, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+    class Meta:
```
mltdll, line 66: class `Meta` should be defined before method definitions.

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+            )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    queryset = Order.objects.all()
+
+    if username:
+        queryset = queryset.filter(user__username=username)
+
+    return queryset.order_by("-user__username")
```
mltdll, line 35: Ordering should not be applied here.
There was a bug that made tests fail on some students' computers. It should be fixed by now. Either way, it shouldn't affect tests on GitHub.

mltdll (changes requested):

mltdll (approved):

settings.py
```diff
@@ -24,5 +24,7 @@
 USE_TZ = False
 
 INSTALLED_APPS = [
-    "db",
+    "db", "django.contrib.auth", "django.contrib.contenttypes"
```
TPolina, line 27: better place apps on the separate lines

TPolina (approved):
Well done!

PR #125, opened 2022-10-27
Solution

OvDima1:


services/order.py
```diff
@@ -0,0 +1,38 @@
...
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+            for ticket in tickets
+        ]
+
+
+def get_orders(username: str = None) -> QuerySet:
+    queryset = Order.objects.all().order_by("-user")
```
mltdll, line 33: `.order_by()` is not needed here. The tests used to fail on some students' computers because of a bug that should now be fixed. The tests should pass on GitHub either way, even if they fail locally.

mltdll (changes requested):

mltdll (approved):

db/models.py
```diff
@@ -45,3 +49,70 @@ class MovieSession(models.Model):
...
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
```
mltdll, line 60: Sorry, one more thing. It's probably better to use `settings.AUTH_USER_MODEL`.

mltdll (changes requested):

mltdll (approved):

Y-Havryliv (approved):
GJ!

PR #127, opened 2022-10-27
Solution!

ketstap162:


db/models.py
```diff
@@ -45,3 +56,69 @@ class MovieSession(models.Model):
...
+                        f"(1, seats_in_row): (1, {max_seats})"
+            })
+
+    def save(
+        self,
+        force_insert: bool = False,
+        force_update: bool = False,
+        using: Any = None,
+        update_fields: Any = None
+    ) -> Ticket:
```
mltdll, line 110: Are you sure this method returns a `Ticket` instantce?

mltdll (changes requested):

db/models.py
ketstap162: My bad, it's mistake

mltdll (approved):

services/user.py
```diff
@@ -0,0 +1,51 @@
+from django.contrib.auth import get_user_model
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = "",
+        first_name: str = "",
+        last_name: str = ""
```
Nattalli, line 9: great idea

Nattalli (approved):

PR #129, opened 2022-10-27
Develop

Kitazuka:


db/models.py
```diff
@@ -45,3 +52,74 @@ class MovieSession(models.Model):
...
+                                      related_name="tickets")
+    order = models.ForeignKey(Order,
+                              on_delete=models.CASCADE,
+                              related_name="tickets")
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(fields=["row", "seat"], name="unique_seat")
```
mltdll, line 86: Shouldn't customers be able to order the same seat number for different sessions? :)

db/models.py
```diff
@@ -45,3 +52,74 @@ class MovieSession(models.Model):
...
+                        f"{self.movie_session.cinema_hall.seats_in_row})"
+                    ]
+                }
+            )
+
+    def save(self,
+             force_insert: bool = False,
+             force_update: bool = False,
+             using: Any = None,
+             update_fields: Any = None) -> Any:
```
mltdll, line 119: I would advice to check what parent class's `.super()` returns and type the same here.

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
+    return [
+        {
+            "row": ticket.row,
+            "seat": ticket.seat
+        }for ticket in Ticket.objects.filter(movie_session_id=movie_session_id)
```
mltdll, line 52: You missed a space here. Anyway, it would more readable to write it this way:
```python
[
    {...}
    for ...
]

File: services/user.py [hunk 4]
```diff
@@ -0,0 +1,51 @@
+from django.contrib.auth import get_user_model
+
+from db.models import User
+
+
+def update_some_info(user: User,
```
mltdll, line 6: Looks more like `personal_info`.

mltdll (changes requested):

db/models.py
```diff
@@ -47,8 +47,12 @@ def __str__(self) -> str:
 
 class MovieSession(models.Model):
     show_time = models.DateTimeField()
-    cinema_hall = models.ForeignKey(to=CinemaHall, on_delete=models.CASCADE)
-    movie = models.ForeignKey(to=Movie, on_delete=models.CASCADE)
+    cinema_hall = models.ForeignKey(
+        to=CinemaHall, on_delete=models.CASCADE, related_name="movie_session"
+    )
+    movie = models.ForeignKey(
```
mltdll, line 53: The name for backwards relation managers should be plural unless it's One-to-One relation.

mltdll (changes requested):

db/models.py
```diff
@@ -40,8 +47,87 @@ def __str__(self) -> str:
...
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+    class Meta:
+        ordering = ["-created_at"]
```
TPolina, line 77: class Meta should be placed before methods

TPolina (changes requested):

TPolina (approved):
Well done!

PR #132, opened 2022-10-27
Solution

skrypka-dmytro:


services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    tickets = Ticket.objects.filter(movie_session__id=movie_session_id)
```
mltdll, line 48: `__id` lookup is not necessary here; use `_id`. `ticket` table already contains it.

mltdll (changes requested):

mltdll (approved):

services/movie_session.py
Nattalli: @mltdll btw, as I know, Danylo told that it is better to use `__id` because sometimes we can create a field with a `_id` suffix and an error may occur. But I use `_id` as well 😅

db/models.py
```diff
@@ -1,8 +1,15 @@
...
 from django.db import models
+from django.db.models import UniqueConstraint
 
 
 class Genre(models.Model):
-    name = models.CharField(max_length=255, unique=True)
+    name = models.CharField(max_length=255)
+
+    class Meta:
+        models.Index(fields=["title"])
```
Nattalli, line 12: why do you make an index for this field? It is not required by the task

db/models.py
```diff
@@ -19,8 +26,13 @@ def __str__(self) -> str:
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
-    actors = models.ManyToManyField(to=Actor)
-    genres = models.ManyToManyField(to=Genre)
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
Nattalli, line 30: add `related_name`
P.S. read the checklist to make it with best practices 😄

Nattalli (changes requested):

db/models.py
skrypka-dmytro: image

services/movie_session.py
mltdll: Got it. Thanks for the information!
It seems that it would be even more flexible to use `__pk` lookup because that way it doesn't depend on the name of the primary key field of the other table.

Nattalli (approved):

PR #139, opened 2022-10-27
Solved the task

Cerne13:


services/user.py
```diff
@@ -0,0 +1,57 @@
...
+
+    if email or first_name or last_name:
+        user.save()
+
+    return user
+
+
+def get_user(user_id: int) -> User:
+    user = User.objects.get(id=user_id)
+    return user
```
YevheniiDavydenko, line 33: you can solve it in one line without user

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+                order_id=order.id,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+        return order
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
+    orders = Order.objects.all().prefetch_related("user")
```
YevheniiDavydenko, line 32: read checklist in readme, you don't need all(). here

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+            order.created_at = date
+            order.save()
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
+                order_id=order.id,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
```
YevheniiDavydenko, line 26: I think that bulk_create will run much faster
also you can use list comprehesions

db/models.py
```diff
@@ -45,3 +55,79 @@ class MovieSession(models.Model):
...
+        self.full_clean()
+
+        return super(Ticket, self).save(
+            force_insert,
+            force_update,
+            using,
+            update_fields
+        )
+
+    class Meta:
```
YevheniiDavydenko, line 127: I don't now, but by my opinion it will be better for readability if you put Meta right after attributes

YevheniiDavydenko (changes requested):

services/order.py
Cerne13: Thank you. Was constantly worried about electricity, so made a couple of stupid mistakes :)

db/models.py
Cerne13: Yes, I agree. Somehow thought that typical order is 'attrs -> methods -> meta'.
Just checked. As far as seen, the best practice is considered 'attrs -> meta -> methods'

services/order.py
Cerne13: Oh, yes. Missed this moment :)
Remembered later, but hadn't time to fix.

zhulich (approved):
GJ!

services/movie_session.py
```diff
@@ -42,3 +42,15 @@ def update_movie_session(
...
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    session = MovieSession.objects.get(id=movie_session_id)
+
+    return [
+        {
+            "row": ticket.row,
+            "seat": ticket.seat
+        }
+        for ticket in session.ticket_set.all()
```
TPolina, line 55: better to add and use related_name

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+            [Ticket(
+                movie_session=MovieSession.objects.get(
+                    id=ticket["movie_session"]
+                ),
+                order_id=order.id,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            ) for ticket in tickets
+            ]
+        )
```
TPolina, line 30: why not iterate trough tickets and create each separately? Your code would be much more neat

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+        order = Order.objects.create(user=user)
+
+        if date:
+            order.created_at = date
+            order.save()
+
+        Ticket.objects.bulk_create(
+            [Ticket(
+                movie_session=MovieSession.objects.get(
+                    id=ticket["movie_session"]
```
TPolina, line 23: you don't need to get `MovieSession` object you can use `movie_session_id`

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+
+    if email or first_name or last_name:
```
TPolina, line 25: you can omit this check

TPolina (changes requested):

services/movie_session.py
Cerne13: Thank you

services/order.py
Cerne13: There were some problems with id during bulk creation. I'll definitely change it back to session id when I change the method to create()

TPolina (approved):
Well done!

PR #140, opened 2022-10-27
Solution

LiudmylaKulinchenko:


db/models.py
```diff
@@ -22,6 +31,9 @@ class Movie(models.Model):
     actors = models.ManyToManyField(to=Actor)
```
mltdll, line 31: It's better to specify `related_name`.

db/models.py
```diff
@@ -45,3 +57,71 @@ class MovieSession(models.Model):
...
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey("MovieSession", on_delete=models.CASCADE)
+    order = models.ForeignKey("Order", on_delete=models.CASCADE)
```
mltdll, line 78: Same here.

db/models.py
```diff
@@ -45,3 +57,71 @@ class MovieSession(models.Model):
...
+                ]}
+            )
+
+    def save(
+            self,
+            force_insert: bool = False,
+            force_update: bool = False,
+            using: Any = None,
+            update_fields: Any = None
+    ) -> Ticket:
```
mltdll, line 116: Does it really return a `Ticket` object? Check the signature of `super().save()`.

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+        date: datetime.date = None
+) -> None:
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
+        order = Order.objects.create(user=user)
+        if date:
+            order.created_at = date
+            order.save()
+        for ticket in tickets:
+            movie_session = MovieSession.objects.get(
```
mltdll, line 22: Use `movie_session_id`, as the [checklist](https://github.com/mate-academy/py-movie-moviesession-cinemahall/blob/master/checklist.md#code-efficiency) suggests.

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+
+    with transaction.atomic():
+        get_user_model().objects.create_user(
+            username=username,
+            password=password,
+            email=email if email else "",
```
mltdll, line 19: You can simplify this expression using `or` operator.

mltdll (changes requested):

db/models.py
```diff
@@ -74,8 +72,16 @@ def __str__(self) -> str:
 
 
 class Ticket(models.Model):
-    movie_session = models.ForeignKey("MovieSession", on_delete=models.CASCADE)
-    order = models.ForeignKey("Order", on_delete=models.CASCADE)
+    movie_session = models.ForeignKey(
+        "MovieSession",
+        on_delete=models.CASCADE,
+        related_name="movie_sessions"
```
mltdll, line 78: Related name works the other way: it should be `tickets` for both foreign keys.

mltdll (changes requested):

mltdll (approved):

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        user = get_user_model().objects.get(username=username)
+        return Order.objects.filter(user=user)
```
TPolina, line 33: you can use `filter(user__username=username)`

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+        last_name: str = None
+) -> None:
+
+    with transaction.atomic():
+        get_user_model().objects.create_user(
+            username=username,
+            password=password,
+            email=email or "",
+            first_name=first_name or "",
+            last_name=last_name or ""
```
TPolina, line 21: strange approach. and when some information was not passed it should be None not ""

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+
+    with transaction.atomic():
```
TPolina, line 15: I am not sure we need transaction here

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = get_user(user_id)
+
+    with transaction.atomic():
```
TPolina, line 39: we don't need transaction here

TPolina (changes requested):

services/user.py
LiudmylaKulinchenko: It doesn't work with None. I use "" because than it will be blank. And these fields cannot be Null. Maybe it is ok to use "" here?

TPolina: You are right about Null. You could use if statements here but seems that your solution is also good

TPolina (approved):
Well done!

PR #143, opened 2022-10-27
Solution

fsocie7y:


db/models.py
```diff
@@ -45,3 +56,78 @@ class MovieSession(models.Model):
...
+                }
+            )
+
+    def save(
+            self,
+            force_insert: bool = False,
+            force_update: bool = False,
+            using: Any = None,
+            update_fields: Any = None
+    ) -> None:
```
Vasyl-Poremchuk, line 122: If you use `return` it will be not a `None` output annotation

Vasyl-Poremchuk (changes requested):

Vasyl-Poremchuk (approved):
Well done!

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+        if date:
+            order.created_at = date
+            order.save()
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                row=ticket["row"],
+                seat=ticket["seat"],
+                movie_session=MovieSession.objects.get(
+                    id=ticket["movie_session"]
```
TPolina, line 26: you don't need to get `MovieSession` object here. You can use `movie_session_id`

TPolina (changes requested):

services/order.py
fsocie7y: yes, but...
image

TPolina: movie_session_id

TPolina (approved):
Well done!

PR #144, opened 2022-10-27
Create models

FranchukS:


db/models.py
```diff
@@ -45,3 +50,73 @@ class MovieSession(models.Model):
...
+                         f"(1, seats_in_row): "
+                         f"(1, {self.movie_session.cinema_hall.seats_in_row})"]
+            })
+
+    def save(
+            self, force_insert: bool = False,
+            force_update: bool = False,
+            using: bool = None,
+            update_fields: bool = None
+    ) -> callable:
```
mltdll, line 105: You can copy type annotation from `.save()` method of the parent class.

mltdll (changes requested):

mltdll (approved):

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+        order = Order.objects.create(
+            user=get_user_model().objects.get(username=username)
+        )
+        if date:
+            order.created_at = date
+            order.save()
+
+        for ticket in tickets:
+            ticket["movie_session_id"] = ticket.pop("movie_session")
+            Ticket.objects.create(order=order, **ticket)
```
TPolina, line 25: it's strange to do these actions

TPolina (changes requested):

services/order.py
FranchukS: 🤷‍♂️ )

TPolina (approved):
Well done!

PR #145, opened 2022-10-27
Solution

pavlo-kapitanets:


db/models.py
```diff
@@ -45,3 +50,72 @@ class MovieSession(models.Model):
...
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        to=settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE)
```
mltdll, line 63: It is important to follow the company style guide. Please, choose one of the following formats and stick to it:
```python
# Break before the opening and after the closing parentheses.
  user = models.ForeignKey(
        to=settings.AUTH_USER_MODEL,
        on_delete=models.CASCADE
)

# No breaks after or before parentheses
  user = models.ForeignKey(to=settings.AUTH_USER_MODEL,
                           on_delete=models.CASCADE)
```

db/models.py
```diff
@@ -45,3 +50,72 @@ class MovieSession(models.Model):
...
+        return str(self.created_at)
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        to=MovieSession,
+        on_delete=models.CASCADE)
+    order = models.ForeignKey(
+        to=Order,
+        on_delete=models.CASCADE)
```
mltdll, line 78: Same thing here.

db/models.py
```diff
@@ -45,3 +50,72 @@ class MovieSession(models.Model):
...
+            force_update: bool = None,
+            using: bool = None,
+            update_fields: bool = None
+    ) -> None:
+        self.full_clean()
+        return super(Ticket, self).save(
+            force_insert,
+            force_update,
+            using,
+            update_fields)
```
mltdll, line 121: And here.

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+        user = get_user_model().objects.get(username=username)
+        order = Order.objects.create(
+            user=user
+        )
+        if date:
+            order.created_at = date
+            order.save()
+
+        for ticket in tickets:
+            movie_session = MovieSession.objects.get(
```
mltdll, line 19: Getting objects here causes unnecessary database load. Use `movie_session_id` argument to pass id directly.

mltdll (changes requested):

db/models.py
pavlo-kapitanets: Thank you for your review! I've chosen the first variant of parentheses style and replaced get method with movie_session_id argument.

mltdll (approved):
GJ!

services/movie_session.py
```diff
@@ -25,10 +25,10 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
...
 
 def update_movie_session(
-    session_id: int,
-    show_time: str = None,
-    movie_id: int = None,
-    cinema_hall_id: int = None,
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None,
```
TPolina, line 31: same here and farther. there should be 4 spaces, not 8

services/user.py
```diff
@@ -0,0 +1,50 @@
...
+def create_user(username: str,
+                password: str,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> None:
+    get_user_model().objects.create_user(
+        username=username,
+        password=password
+    )
+    user = get_user_model().objects.get(username=username)
```
TPolina, line 14: `create_user` returns created object. You don't need to make get query separately

TPolina (changes requested):

TPolina (approved):
Well done!

PR #146, opened 2022-10-27
Solution

Nikyxa:


db/models.py
```diff
@@ -45,3 +50,80 @@ class MovieSession(models.Model):
...
+class User(AbstractUser):
+    pass
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        "MovieSession",
+        on_delete=models.CASCADE
+    )
+    order = models.ForeignKey("Order", on_delete=models.CASCADE)
```
mltdll, line 78: It might be better to add `related_name` to the foreign key fields.

db/models.py
```diff
@@ -45,3 +50,80 @@ class MovieSession(models.Model):
...
+            force_update,
+            using,
+            update_field
+        )
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_constraints"
```
mltdll, line 121: This name is not descriptive. I think it's better to chose something that reflects that it's a unique seat for a given session.

mltdll (approved):

mltdll (changes requested):
Oops, wrong button.

mltdll (approved):

db/models.py
```diff
@@ -41,7 +46,93 @@ def __str__(self) -> str:
...
+            update_field
+        )
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket"
+            )
+        ]
```
TPolina, line 132: class Meta should be placed before methods

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+                row=ticket["row"],
+                seat=ticket["seat"],
+                order=order
+            )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        user = get_user_model().objects.get(username=username)
+        return Order.objects.filter(user=user)
```
TPolina, line 32: you can use `filter(user__username=username)`

TPolina (changes requested):

PR #147, opened 2022-10-28
Solution

pazyn404:


db/models.py
```diff
@@ -40,8 +48,95 @@ def __str__(self) -> str:
...
+    )
+    movie = models.ForeignKey(to="Movie", on_delete=models.CASCADE)
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(to="User", on_delete=models.CASCADE)
```
mltdll, line 62: Use `settings.AUTH_USER_MODEL` to refer to the user model in definitions of other models.

db/models.py
```diff
@@ -40,8 +48,95 @@ def __str__(self) -> str:
...
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        to="MovieSession",
+        on_delete=models.CASCADE,
+        related_name="movie_sessions",
+    )
+    order = models.ForeignKey(
+        to="Order", on_delete=models.CASCADE, related_name="orders"
```
mltdll, line 78: The related name represents the name of the attribute of the related model through which you can access related `Tickets`. Therefore, it should be `tickets`, not the other way around.

db/models.py
```diff
@@ -40,8 +48,95 @@ def __str__(self) -> str:
 
 class MovieSession(models.Model):
     show_time = models.DateTimeField()
-    cinema_hall = models.ForeignKey(to=CinemaHall, on_delete=models.CASCADE)
-    movie = models.ForeignKey(to=Movie, on_delete=models.CASCADE)
+    cinema_hall = models.ForeignKey(
+        to="CinemaHall", on_delete=models.CASCADE, related_name="cinema_halls"
```
mltdll, line 52: Here's the same issue.
By the way, why set `related_name` only to some of the fields? It seems like an inconsistency that will make your models confusing to use.

db/models.py
```diff
@@ -40,8 +48,95 @@ def __str__(self) -> str:
...
+                f"{self.movie_session.show_time} "
+                f"(row: {self.row}, seat: {self.seat})")
+
+    def save(
+        self,
+        force_insert: bool = False,
+        force_update: bool = False,
+        using: bool = None,
+        update_fields: bool = None,
+    ) -> Ticket:
```
mltdll, line 106: This method's type annotation is incorrect. You can refer to the signature of `super().save()` and copy it.

db/models.py
```diff
@@ -40,8 +48,95 @@ def __str__(self) -> str:
...
+        self.full_clean()
+        return super(Ticket, self).save(
+            force_insert, force_update, using, update_fields
+        )
+
+    def clean(self) -> None:
+        cinema_hall = self.movie_session.cinema_hall
+
+        row_criteria = 1 <= self.row <= cinema_hall.rows
+        seat_criteria = 1 <= self.seat <= cinema_hall.seats_in_row
```
mltdll, line 116: In my opinion, these variables don't seem to make the code easier to read. Maybe even the opposite.

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+                row=ticket["row"],
+                seat=ticket["seat"],
+                movie_session_id=ticket["movie_session"],
+                order=new_order,
+            )
+            for ticket in tickets
+        ]
+
+        for ticket in tickets:
+            ticket.save()
```
mltdll, line 30: It might be better to create tickets in place using `.objects.create()` to avoid creating the second loop.

services/user.py
```diff
@@ -0,0 +1,47 @@
...
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None,
+) -> User:
+    return get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        email=email if email else "",
+        first_name=first_name if first_name else "",
+        last_name=last_name if last_name else "",
```
mltdll, line 22: The `email` field can be `None`, the other two expressions may be simplified using `or` operators. 
Alternatively, it's possible to set default names to those arguments to `""` even though it contradicts instructions. :)

mltdll (changes requested):

db/models.py
```diff
@@ -93,29 +103,28 @@ class Meta:
...
+            f"{self.movie_session.show_time} "
+            f"(row: {self.row}, seat: {self.seat})"
+        )
 
     def save(
         self,
         force_insert: bool = False,
         force_update: bool = False,
         using: bool = None,
         update_fields: bool = None,
```
mltdll, line 117: Still not quite correct. :)

services/order.py
```diff
@@ -16,18 +16,13 @@ def create_order(
...
                 seat=ticket["seat"],
                 movie_session_id=ticket["movie_session"],
                 order=new_order,
-            )
-            for ticket in tickets
-        ]
-
-        for ticket in tickets:
-            ticket.save()
+            ).save()
```
mltdll, line 25: I can't see anything wrong with this implementation, but I would have used `Ticket.objects.create()` because this one is definitely fine.

mltdll (changes requested):

mltdll (approved):
GJ!

Nattalli (approved):

PR #148, opened 2022-10-28
Solution

misha-doroshenko:


db/models.py
```diff
@@ -45,3 +55,68 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey("User", on_delete=models.CASCADE)
```
mltdll, line 62: It might be better to specify `related_name` for all the foreign key fields.

db/models.py
```diff
@@ -45,3 +55,68 @@ class MovieSession(models.Model):
...
+                          f"{self.movie_session.cinema_hall.seats_in_row})"]}
+            )
+
+    def save(
+            self,
+            force_insert: bool = False,
+            force_update: bool = False,
+            using: Any = None,
+            update_fields: Any = None
+    ) -> Any:
```
mltdll, line 105: I'd suggest to check what `super().save()` method returns.

db/models.py
```diff
@@ -45,3 +55,68 @@ class MovieSession(models.Model):
...
+            force_update,
+            using,
+            update_fields
+        )
+
+    def __str__(self) -> str:
+        return (
+            f"{self.movie_session.movie.title} "
+            f"{self.movie_session.show_time} "
+            f"(row: {self.row}, seat: {self.seat})")
```
mltdll, line 118: The closing parenthesis should be on the next line.

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+        last_name: str = None
+
+) -> None:
+    user_to_update = get_user(user_id)
+
+    if username:
+        user_to_update.username = username
+    if password:
+        user_to_update.set_password(raw_password=password)
+    if email:
```
mltdll, line 48: I'd suggest to use `is not None` for the last three fields, since they can be blank.

db/models.py
```diff
@@ -24,8 +24,8 @@ def __str__(self) -> str:
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
-    actors = models.ManyToManyField(to=Actor)
-    genres = models.ManyToManyField(to=Genre)
+    actors = models.ManyToManyField(to=Actor, related_name="actors")
+    genres = models.ManyToManyField(to=Genre, related_name="genres")
```
mltdll, line 28: Related names work the other way: it should be "movies" here.

mltdll (changes requested):

mltdll (approved):

db/models.py
```diff
@@ -40,8 +50,94 @@ def __str__(self) -> str:
...
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        User,
```
TPolina, line 75: you should refer to AUTH_USER_MODEL

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+from datetime import datetime
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> None:
+    with transaction.atomic():
+        order = Order.objects.create(user=User.objects.get(username=username))
```
TPolina, line 14: You should not refer to User directly

TPolina (changes requested):

db/models.py
TPolina: You didn't fix this

TPolina, line 75: you should refer to AUTH_USER_MODEL here

TPolina (changes requested):

TPolina (approved):
Well done!

PR #149, opened 2022-10-28
Added migrations

alexzubkoff:


services/user.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def create_user(
+    username: str,
+    password: str,
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None
+) -> QuerySet:
+    user = User.objects.create_user(
```
Vasyl-Poremchuk, line 13: I think `get_user_model()` should also be used here

services/user.py
```diff
@@ -0,0 +1,53 @@
...
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
+
+
+def get_user(user_id: int) -> QuerySet:
+    return User.objects.get(id=user_id)
```
Vasyl-Poremchuk, line 28: The same here

services/user.py
```diff
@@ -0,0 +1,53 @@
...
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+
+) -> User:
+    user = User.objects.get(id=user_id)
```
Vasyl-Poremchuk, line 40: And here

Vasyl-Poremchuk (changes requested):

services/user.py
alexzubkoff: Thanks! You mean that.
If you reference [User] directly (for example, by referring to it in a foreign key), your code will not work in projects where the [AUTH_USER_MODEL] setting has been changed to a different user model.

get_user_model() you should reference the user model using django.contrib.auth.get_user_model(). This method will return the currently active user model – the custom user model if one is specified, or [User]nusing the [AUTH_USER_MODEL]

alexzubkoff: As i understood from django doc we can use get_user_model() or author = models.ForeignKey(
        settings.AUTH_USER_MODEL,
        on_delete=models.CASCADE,
    )
 Generally speaking, it’s easiest to refer to the user model with the AUTH_USER_MODEL setting in code that’s executed at import time, however, it’s also possible to call get_user_model(). Thanks for proposition!

services/user.py
Vasyl-Poremchuk: You should write `get_user_model().objects.create_user(...)` as this is the best practice you can see in the `checklist.md` file. And the answer to this is also in your post)

Vasyl-Poremchuk (changes requested):

alexzubkoff:
Ok! Thanks!

Сб, 29 окт. 2022 г. в 09:56, Vasyl Poremchuk ***@***.***>:

Vasyl-Poremchuk (approved):

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+
+        if date:
+            order.created_at = date
+            order.save()
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                row=ticket["row"],
+                seat=ticket["seat"],
+                movie_session=MovieSession.objects.get(
```
Y-Havryliv, line 24: since you have `movie_session_id` in tickets, it's better to use it and don't make unnecessary queries

services/user.py
```diff
@@ -0,0 +1,52 @@
...
+from db.models import User
+
+
+def create_user(
+    username: str,
+    password: str,
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None
+) -> QuerySet:
```
Y-Havryliv, line 11: it doesn't return `QuerySet`

services/user.py
Y-Havryliv: I agree, use `get_user_model()`

services/user.py
```diff
@@ -0,0 +1,52 @@
...
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
+
+
+def get_user(user_id: int) -> QuerySet:
```
Y-Havryliv, line 26: it doesn't return `QuerySet`

services/user.py
Y-Havryliv: I agree

services/user.py
Y-Havryliv: use `get_user_model()`

services/user.py
```diff
@@ -0,0 +1,52 @@
...
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+
+) -> User:
+    user = User.objects.get(id=user_id)
+    if user:
```
Y-Havryliv, line 40: this check is unnescessary

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Nice!

PR #150, opened 2022-10-30
'Solution'

Ilyakson:


db/models.py
```diff
@@ -45,3 +50,73 @@ class MovieSession(models.Model):
...
+    user = models.ForeignKey(
+        to=settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE
+    )
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+    class Meta:
+        ordering = ["-created_at"]
```
mltdll, line 66: Class `Meta` should be defined before the methods.

services/movie_session.py
```diff
@@ -42,3 +44,8 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    ticket = Ticket.objects.filter(movie_session=movie_session_id)
```
mltdll, line 50: It's `tickets`, not `ticket`.

mltdll (changes requested):

mltdll (approved):

services/movie.py
```diff
@@ -1,14 +1,19 @@
...
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None,
+        title: str = None
```
TPolina, line 10: what did you add additional spaces for?

services/movie.py
```diff
@@ -23,18 +28,20 @@ def get_movie_by_id(movie_id: int) -> Movie:
...
 
 def create_movie(
-    movie_title: str,
-    movie_description: str,
-    genres_ids: list = None,
-    actors_ids: list = None,
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list = None,
+        actors_ids: list = None,
```
TPolina, line 34: same here and farther. there should be 4 spaces not 8

TPolina (changes requested):

services/movie.py
Ilyakson: These are the default settings in python

services/movie.py
Ilyakson: So that there are no differences in the code

TPolina: This is code style. Indentation in Python is 4 spaces. We can use 8 in function declaration to distinguish parameters. But here it's not necessary because they will be distinguished with 4 spaces as well

TPolina (changes requested):

TPolina (approved):

PR #151, opened 2022-10-31
Solution

serviza:


db/models.py
```diff
@@ -40,8 +43,88 @@ def __str__(self) -> str:
...
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        to=User, on_delete=models.CASCADE, related_name="orders"
+    )
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+    class Meta:
```
mltdll, line 70: `Meta` should be defined before the methods.

db/models.py
```diff
@@ -40,8 +43,88 @@ def __str__(self) -> str:
...
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        to=User, on_delete=models.CASCADE, related_name="orders"
```
mltdll, line 64: It's better to use `settings.AUTH_USER_MODEL`.

services/movie_session.py
```diff
@@ -42,3 +41,12 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    ticket = Ticket.objects.filter(movie_session__id=movie_session_id)
```
mltdll, line 47: This variable should probably be called `tickets`.

services/movie_session.py
```diff
@@ -42,3 +41,12 @@ def update_movie_session(
...
+
+    return [
+        {"row": movie_session.row, "seat": movie_session.seat}
+        for movie_session in ticket
```
mltdll, line 51: And `movie_session` looks like `ticket` to me.

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+                row=ticket["row"],
+                seat=ticket["seat"],
+                order=order
+            )
+            for ticket in tickets
+        ]
+
+
+def get_orders(username: str = None) -> QuerySet:
+    orders = Order.objects.order_by("-user")
```
mltdll, line 30: Ordering is not needed here. The tests should pass without it on GitHub even if they fail locally; there was a bug in tests that is now fixed.

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+
+        [
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
+                row=ticket["row"],
+                seat=ticket["seat"],
+                order=order
+            )
+            for ticket in tickets
+        ]
```
mltdll, line 26: I personally don't like the usage of list comprehension here, but it's probably a question of preference.

mltdll (changes requested):

services/order.py
serviza: Thanks's for review)
Fixed.

db/models.py
mltdll: You didn't fix this one.

services/order.py
mltdll: And this one.

mltdll (changes requested):

db/models.py
serviza: Maybe we don't understand each other.

services/order.py
serviza: Yes. I tried and this works)

db/models.py
mltdll: `User` model should not be referred to directly. You can read more about this [here](https://docs.djangoproject.com/en/4.1/topics/auth/customizing/#referencing-the-user-model).

> When you define a foreign key or many-to-many relations to the user model, you should specify the custom model using the `AUTH_USER_MODEL` setting.

mltdll (changes requested):

serviza:
Ok, I use this settings. But I just looked at your implementation. Why didn't you do it?

mltdll (approved):

mltdll:
> Why didn't you do it?
I probably forgot about it in this task. But I'm pretty sure that I have used it in all the following ones. :)

PR #156, opened 2022-11-14
Solution

hai-le-di:


services/order.py
```diff
@@ -0,0 +1,34 @@
...
+        for ticket in tickets:
+            Ticket.objects.create(
+                order=order,
+                movie_session=get_movie_session_by_id(ticket["movie_session"]),
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+        if date:
+            order.created_at = date
+            order.save()
```
Abnormaltype, line 25: it should be done before creating Tickets

Abnormaltype (changes requested):

PR #170, opened 2022-12-14
Solution added

Ihor-Horiainov:


services/user.py
```diff
@@ -0,0 +1,47 @@
...
+        user_id: int,
+        username: str = None,
+        password: str = False,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user_to_update = get_user_model().objects.get(id=user_id)
+
+    if username:
```
IvanRamyk, line 32:
```suggestion
    if username is not None:
```
please, fix everywhere

IvanRamyk: due to zen of Python 
"Explicit is better than implicit."

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #174, opened 2022-12-22
solution

taras-zelinskyi:


services/user.py
```diff
@@ -0,0 +1,51 @@
...
+        password=password
+    )
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
```
Abnormaltype, line 24: unnecessary `return`

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+    if password:
+        user.set_password(f"{password}")
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
```
Abnormaltype, line 51: the same

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> User:
+    user = get_user_model().objects.get(id=user_id)
```
Abnormaltype, line 39: why so you need `get_user`?

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+    if username:
+        user.username = username
+    if password:
+        user.set_password(f"{password}")
```
Abnormaltype, line 43:
```suggestion
        user.set_password(password)
```

Abnormaltype (changes requested):

services/user.py
taras-zelinskyi: Created new pull request

PR #175, opened 2022-12-22
solution

viktory-koroliova:


services/user.py
```diff
@@ -0,0 +1,41 @@
...
+
+def create_user(
+        username: str,
+        password: str,
+        ** info
+) -> None:
+    get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        ** info
```
Abnormaltype, line 14: why do you need this?

services/user.py
```diff
@@ -0,0 +1,41 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None,
+        password: str = None
+) -> None:
+    user_to_update = get_user_model().objects.get(id=user_id)
```
Abnormaltype, line 30: why do you need `get_user`?

Abnormaltype (changes requested):

services/user.py
Abnormaltype: what about it?

Abnormaltype (changes requested):

services/user.py
viktory-koroliova: What exactly do I have to fix?

Abnormaltype: `** info`

Serhii-Leonenko (approved):
Good job!

PR #207, opened 2023-02-13
Solution

MarharytaSovpenko:


services/movie_session.py
```diff
@@ -1,22 +1,28 @@
...
     )
 
 
-def get_movies_sessions(session_date: str = None) -> QuerySet:
+def get_movies_sessions(session_date: Optional[str] = None) -> QuerySet:
     queryset = MovieSession.objects.all()
+
     if session_date:
-        queryset = queryset.filter(show_time__date=session_date)
+        queryset = MovieSession.objects.filter(show_time__date=session_date)
```
ArsenPidhoretskyi, line 25: Why did you change this? As for me previous was okay

ArsenPidhoretskyi (changes requested):

services/movie_session.py
MarharytaSovpenko: I didn't mean to replace exactly this fragment) just replaced all moviesession.py with my code to avoid mistakes in the annotation concerning Optional:) I'll fix it right now.

services/movie.py
```diff
@@ -19,22 +26,26 @@ def get_movies(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
 
-    return movie
+    with transaction.atomic():
```
Bohdan-Salamakha, line 39: use @transaction.atomic() decorator for it
u don't have an additional logic before

Bohdan-Salamakha (changes requested):
v1

services/movie.py
MarharytaSovpenko: Alright, thank you!

services/movie.py
```diff
@@ -19,22 +26,26 @@ def get_movies(
...
-    genres_ids: list = None,
-    actors_ids: list = None,
+        movie_title: str,
+        movie_description: str,
+        genres_ids: Optional[list[int]] = None,
+        actors_ids: Optional[list[int]] = None
 ) -> Movie:
-    movie = Movie.objects.create(
+
+    new_movie = Movie.objects.create(
```
ArsenPidhoretskyi, line 40: No need to rename to new_movie
But okay

services/movie_session.py
```diff
@@ -25,20 +31,26 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
 
 
 def update_movie_session(
-    session_id: int,
-    show_time: str = None,
-    movie_id: int = None,
-    cinema_hall_id: int = None,
+        session_id: int,
+        show_time: datetime = None,
```
ArsenPidhoretskyi, line 35: Optional[datetime]

ArsenPidhoretskyi (changes requested):

services/movie_session.py
MarharytaSovpenko: > Optional[datetime]

optional er
I face this error( how to deal with it?

ArsenPidhoretskyi (approved):
Well done!

PR #208, opened 2023-02-14
Solution

evgenijmartynuk07:


services/movie.py
```diff
@@ -1,14 +1,17 @@
 from django.db.models import QuerySet
-
+from django.db import transaction
 from db.models import Movie
 
 
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
```
ArsenPidhoretskyi, line 8: Change to Optional[list[int]]
Because by default it accepts None
Check and do the same in other places

services/user.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def get_user(user_id: int) -> User:
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(user_id: int, **kwargs) -> None:
+    user = get_user(user_id)
+    for key, value in kwargs.items():
+        if key == "password":
```
ArsenPidhoretskyi, line 27: Better to move to separate value in function parameters

ArsenPidhoretskyi (changes requested):

db/models.py
```diff
@@ -46,3 +63,49 @@ class MovieSession(models.Model):
...
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.movie_session.movie.title} " \
+               f"{self.movie_session.show_time} " \
+               f"(row: {self.row}, seat: {self.seat})"
```
katevasilenko, line 77: Please use parentheses to avoid \ symbols

db/models.py
```diff
@@ -46,3 +63,49 @@ class MovieSession(models.Model):
...
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket"
+            )
+        ]
+
+    def clean(self) -> None:
+        if self.row > self.movie_session.cinema_hall.rows:
```
katevasilenko, line 88: And what if it negative or zero?

db/models.py
```diff
@@ -46,3 +63,49 @@ class MovieSession(models.Model):
...
+            raise ValidationError(
+                {
+                    "row":
+                        [
+                            "row number must be in available range: "
+                            "(1, rows): (1, 10)"
+                        ]
+                }
+            )
+        if self.seat > self.movie_session.cinema_hall.seats_in_row:
```
katevasilenko, line 98: The same question is here

db/models.py
```diff
@@ -46,3 +63,49 @@ class MovieSession(models.Model):
...
+        ]
+
+    def clean(self) -> None:
+        if self.row > self.movie_session.cinema_hall.rows:
+            raise ValidationError(
+                {
+                    "row":
+                        [
+                            "row number must be in available range: "
+                            "(1, rows): (1, 10)"
```
katevasilenko, line 94: Why 10? Use variable to set range

db/models.py
```diff
@@ -46,3 +63,49 @@ class MovieSession(models.Model):
...
+                        ]
+                }
+            )
+        if self.seat > self.movie_session.cinema_hall.seats_in_row:
+            raise ValidationError(
+                {
+                    "seat":
+                        [
+                            "seat number must be in available range: "
+                            "(1, seats_in_row): (1, 12)"
```
katevasilenko, line 104: Use variable to set range

katevasilenko (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #211, opened 2023-02-20
solution

xpanzhar:


services/movie.py
```diff
@@ -1,11 +1,13 @@
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+        genres_ids: list[int] = None,
```
ArsenPidhoretskyi, line 8: if default is None
it already optional parameter 
change typing in other places the same

ArsenPidhoretskyi (changes requested):

services/movie.py
AlexanderTsvirko: its done, please check

ArsenPidhoretskyi: Not done 
image
The same in other places

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job!

PR #212, opened 2023-02-21
'Solution'

Oleksiy-Liubchenko:


services/movie.py
```diff
@@ -28,13 +33,15 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
 
-    return movie
+    with transaction.atomic():
```
Bohdan-Salamakha, line 37: use @transaction.atomic() decorator

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+from django.db.models import QuerySet
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> Order:
+
+    with transaction.atomic():
```
Bohdan-Salamakha, line 15: use @transaction.atomic() decoreator

services/movie_session.py
```diff
@@ -42,3 +42,15 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict[str, int]]:
+    movie_session = MovieSession.objects.get(id=movie_session_id)
```
Bohdan-Salamakha, line 48: use get_object_or_404 to avoid Error

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+        user = get_user_model().objects.get(username=username)
```
Bohdan-Salamakha, line 16: use get_object_or_404 to avoid Error

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+        order = Order.objects.create(user=user)
+
+        for ticket in tickets:
+            if ticket["movie_session"]:
+                movie_session = MovieSession.objects.get(
```
Bohdan-Salamakha, line 21: use get_object_or_404 to avoid Error

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
```
Bohdan-Salamakha, line 26: use get_object_or_404 to avoid Error

Bohdan-Salamakha (changes requested):
v1

services/order.py
Oleksiy-Liubchenko: With decorator it will work the same, but pros is in less rows of code ?

ArsenPidhoretskyi (approved):
Well done!

PR #213, opened 2023-02-21
Solution

alina-boichenko:


services/user.py
```diff
@@ -0,0 +1,44 @@
...
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None
+) -> None:
+    user = User.objects.create_user(
+        username=username,
+        password=password
+    )
+
+    if email:
```
Grayder0152, line 16: More correct `if email is not None:`

services/user.py
```diff
@@ -0,0 +1,44 @@
...
+        user.email = email
+    if first_name:
```
Grayder0152, line 18: the same as above

services/user.py
```diff
@@ -0,0 +1,44 @@
...
+        user.first_name = first_name
+    if last_name:
```
Grayder0152, line 20: the same as above

services/user.py
```diff
@@ -0,0 +1,44 @@
...
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
+
+
+def update_user(
+    user_id: int, username: str = None, password: str = None,
+    email: str = None, first_name: str = None, last_name: str = None
+) -> None:
+    user = get_user(user_id)
+    if username:
```
Grayder0152, line 34: the same as above

services/user.py
```diff
@@ -0,0 +1,44 @@
+from db.models import User
+
+
+def create_user(
+    username: str,
+    password: str,
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None
```
Grayder0152, line 9: `Oprional[str]` or str | None - correct typing for that args

PR #214, opened 2023-02-21
Solution

alina-boichenko:


services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
ArsenPidhoretskyi, line 10: Is is Optional[str] because it accepts None as default

services/order.py
```diff
@@ -0,0 +1,32 @@
+from django.contrib.auth import get_user_model
+from django.db import transaction
+
+from db.models import Order, Ticket
+
+
+def create_order(
+    tickets: list[dict],
+    username: str,
+    date: str = None
```
ArsenPidhoretskyi, line 10: Is is Optional[str] because it accepts None as default

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+
+def get_orders(username: str = None) -> Order:
```
ArsenPidhoretskyi, line 29: Is is Optional[str] because it accepts None as default

services/user.py
```diff
@@ -0,0 +1,44 @@
+from db.models import User
+
+
+def create_user(
+    username: str,
+    password: str,
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None
```
ArsenPidhoretskyi, line 9: Is is Optional[str] because it accepts None as default

services/user.py
```diff
@@ -0,0 +1,44 @@
...
+    user.save()
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
+
+
+def update_user(
+    user_id: int, username: str = None, password: str = None,
+    email: str = None, first_name: str = None, last_name: str = None
```
ArsenPidhoretskyi, line 31: Fix typing due to Optional

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #220, opened 2023-03-03
'Solution'

timyr4iq:


services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
ArsenPidhoretskyi, line 10: it is Optional[str]

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket, User
+from datetime import datetime
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
```
ArsenPidhoretskyi, line 11: Optional[datetime]

ArsenPidhoretskyi: Fix in other places with all types

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done! Usage `is not None` cool practice

PR #221, opened 2023-03-03
'Solution'

rakamakaphone:


db/models.py
```diff
@@ -46,3 +54,65 @@ class MovieSession(models.Model):
...
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(settings.AUTH_USER_MODEL,
+                             on_delete=models.CASCADE,
+                             null=True
+                             )
+
+    def __str__(self) -> str:
```
Bohdan-Salamakha, line 66: add typehint

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
ArsenPidhoretskyi, line 10: Optional[str] here
If default parameter is None it must be Optional
Fix in all places(only your code)

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Good job!

PR #226, opened 2023-03-07
Dev roman senyshyn py tickets and orders django

AndereLion:


services/order.py
```diff
@@ -0,0 +1,38 @@
...
+            )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    orders = Order.objects.all()
+
+    if username:
+        orders = orders.filter(user__username=username)
+
+    return orders
```
WDemidenko, line 38: you can make it in 2 lines, with ternary operator

services/user.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+
+    if email:
+        user.email = email
+
+    if first_name:
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
```
WDemidenko, line 25: You can replace it with **kwargs, and then unpack kwargs into create_user, so if there was email, first or last names, it will create with correct value

services/user.py
```diff
@@ -0,0 +1,59 @@
...
+        user.username = username
+
+    if email:
+        user.email = email
+
+    if first_name:
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
```
WDemidenko, line 57: I think its not a bad solution, but try to use settatr, it would remove repeated code and checking.

WDemidenko (changes requested):

services/order.py
AndereLion: Fixed

services/user.py
AndereLion: Done

services/user.py
AndereLion: Done

services/user.py
```diff
@@ -3,26 +3,20 @@
...
 
-    if first_name:
-        user.first_name = first_name
+    if kwargs.get("first_name"):
+        user.first_name = kwargs["first_name"]
 
-    if last_name:
-        user.last_name = last_name
+    if kwargs.get("last_name"):
+        user.last_name = kwargs["last_name"]
```
WDemidenko, line 19: this is redundant, create_user will work like this
```
get_user_model().objects.create_user(
        username=username,
        password=password,
        **kwargs
    )
```

WDemidenko (approved):
Good job!

services/user.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+
+    if kwargs.get("email"):
+        user.email = kwargs["email"]
+
+    if kwargs.get("first_name"):
+        user.first_name = kwargs["first_name"]
+
+    if kwargs.get("last_name"):
+        user.last_name = kwargs["last_name"]
```
wQuelS, line 19: if you're using `**kwargs` you can create user without those `if statements` using unpacking

services/user.py
```diff
@@ -0,0 +1,38 @@
...
+
+    user.save()
+
+
+def get_user(user_id: int) -> User:
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(user_id: int, **kwargs) -> None:
+    user = get_user_model().objects.get(id=user_id)
```
wQuelS, line 29: why don't just use `get_user()` method which you previously created?

wQuelS (changes requested):

services/user.py
AndereLion: Fixed all  ;)

YuriiKindrat (approved):

PR #227, opened 2023-03-07
Solution

SlavikSherbak:


db/models.py
```diff
@@ -46,3 +56,74 @@ class MovieSession(models.Model):
...
+    order = models.ForeignKey(
+        "Order",
+        on_delete=models.CASCADE,
+        related_name="tickets"
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.movie_session.movie.title} " \
```
vasylhnatiuk, line 87: use () instead / for long str

db/models.py
```diff
@@ -46,3 +56,74 @@ class MovieSession(models.Model):
...
+                "seat": f"seat number must be in available range: "
+                        f"(1, seats_in_row): "
+                        f"(1, {self.movie_session.cinema_hall.seats_in_row})"
+            })
+
+    def save(
+            self,
+            force_insert: bool = False,
+            force_update: bool = False,
+            using: Any = None,
```
vasylhnatiuk, line 116: add a more accurate annotation, don't use Any here

vasylhnatiuk (changes requested):

db/models.py
SlavikSherbak: image

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
ArsenPidhoretskyi, line 10: It accepts Optional[str] because by default it accepts None
Fix in other places(only your code)

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #228, opened 2023-03-08
Solution

WDemidenko:


db/models.py
```diff
@@ -46,3 +57,69 @@ class MovieSession(models.Model):
...
+    def clean(self) -> None:
+        if not self.row <= self.movie_session.cinema_hall.rows:
+            raise ValidationError({
+                "row": (f"row number must be in available range: (1, rows): "
+                        f"(1, {self.movie_session.cinema_hall.rows})")
+            })
+        if not self.seat <= self.movie_session.cinema_hall.seats_in_row:
+            raise ValidationError({
+                "seat": (f"seat number must be in available range: "
+                         f"(1, seats_in_row): "
```
MaybeTI, line 108: no need f here

MaybeTI (approved):

services/user.py
```diff
@@ -0,0 +1,39 @@
...
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = get_user_model().objects.get(id=user_id)
+
+    arguments = ["username", "password", "email", "first_name", "last_name"]
+    for argument in arguments:
+        if locals()[argument]:
```
wQuelS, line 34: `locals()` are usually used in some sort of debugging, but not in real code. You could use `**kwargs` as you did in the previous function

wQuelS (changes requested):

db/models.py
```diff
@@ -46,3 +57,69 @@ class MovieSession(models.Model):
...
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket"
+            )
+        ]
+
+    def __str__(self) -> str:
+        return (f"{self.movie_session.movie.title} "
+                f"{self.movie_session.show_time.strftime('%Y-%m-%d %H:%M:%S')}"
```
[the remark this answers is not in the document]
YuriiKindrat: what about this comment ?

YuriiKindrat (changes requested):

PR #234, opened 2023-03-12
Solution

SergiiMachulin:


ZabFTFT (approved):
Well done!

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+        for ticket_data in tickets:
+            row, seat, movie_session_id = ticket_data.values()
+            movie_session = MovieSession.objects.get(id=movie_session_id)
+            Ticket.objects.create(
+                order=new_order,
+                row=row,
+                seat=seat,
+                movie_session=movie_session,
+            )
+        return new_order
```
ZabFTFT, line 31: You don't need to get movie_session here, in Ticket parameters just use movie_session_id=use movie_session_id and you will have same result.Tried by myself, it worked =)

SergiiMachulin: Thank you!)

services/movie_session.py
```diff
@@ -42,3 +42,13 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    movie_session = MovieSession.objects.get(id=movie_session_id)
+    taken_seats = list(
+        movie_session.ticket_set.all().order_by(
```
YuriiKindrat, line 50: no need to use all before order by

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None,
+) -> Order:
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
+        new_order = Order.objects.create(user=user)
+        if date:
+            created_at_format = datetime.strptime(date, "%Y-%m-%d %H:%M")
```
YuriiKindrat, line 18: no need to use strftime here

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+        return new_order
+
+
+def get_orders(username: str = None) -> QuerySet:
+    queryset = Order.objects.all()
+
+    if username:
+        user = get_user_model().objects.get(username=username)
+        return queryset.filter(user=user)
+    return queryset
```
YuriiKindrat, line 39: no need to get user model here, use lookups instead, also u can use only 1 return here

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -46,3 +56,78 @@ class MovieSession(models.Model):
...
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE
+    )
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return str(self.created_at)[:19]
```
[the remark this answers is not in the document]
SergiiMachulin: I did it because of output example in README.md. There, we have output without milliseconds, and without slicing or strptime method I don't know how to achieve this? Or is this unnecessary?

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+
+        for ticket_data in tickets:
+            row, seat, movie_session_id = ticket_data.values()
+            Ticket.objects.create(
+                order=new_order,
+                row=row,
+                seat=seat,
+                movie_session_id=movie_session_id,
+            )
+        return new_order
```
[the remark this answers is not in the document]
SergiiMachulin: Try to show different approach of implementation)

services/movie_session.py
```diff
@@ -42,3 +42,13 @@ def update_movie_session(
...
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    movie_session = MovieSession.objects.get(id=movie_session_id)
+    taken_seats = list(
+        movie_session.tickets.order_by(
+            "row", "seat"
+        ).values("row", "seat")
+    )
+    return taken_seats
```
YuriiKindrat, line 54: try to simplify this method u can use ticket model here

services/user.py
```diff
@@ -0,0 +1,52 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None,
+) -> None:
+    user_to_update = get_user_model().objects.get(id=user_id)
```
YuriiKindrat, line 39: u can use get_user func here)

YuriiKindrat (changes requested):

services/movie_session.py
```diff
@@ -45,10 +45,8 @@ def delete_movie_session_by_id(session_id: int) -> None:
...
 
 def get_taken_seats(movie_session_id: int) -> list[dict]:
-    movie_session = MovieSession.objects.get(id=movie_session_id)
-    taken_seats = list(
-        movie_session.tickets.order_by(
-            "row", "seat"
-        ).values("row", "seat")
+    return list(
+        Ticket.objects.filter(movie_session_id=movie_session_id)
+        .order_by("row", "seat")
```
YuriiKindrat, line 50: why do u need order_by here?

YuriiKindrat (approved):

services/movie_session.py
SergiiMachulin: To have sorting to only by row, but also by seat as in example of README.md)

PR #238, opened 2023-03-13
Solution

Vitalii-Khmura:


services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
olena-hul, line 10:
```suggestion
    title: Optional[str] = None
```

services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    session = MovieSession.objects.get(id=movie_session_id)
```
olena-hul, line 48: You can use filter and prefetch related here

olena-hul (changes requested):

services/movie_session.py
Bohdan-Salamakha: not fixed

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+
+from db.models import Order, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: Optional[str] = None
+) -> None:
+    with transaction.atomic():
```
Bohdan-Salamakha, line 14: use @transaction.atomic() decorator

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #239, opened 2023-03-14
Solution

ZabFTFT:


yura-hudzovskyi (changes requested):

services/movie.py
```diff
@@ -15,13 +17,17 @@ def get_movies(
     if actors_ids:
         queryset = queryset.filter(actors__id__in=actors_ids)
 
+    if title:
+        queryset = queryset.filter(title__contains=title)
+
```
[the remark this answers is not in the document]
ZabFTFT: Thank you! I misunderstood this.

db/models.py
```diff
@@ -46,3 +54,76 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.CASCADE
```
wQuelS, line 62: add `related_name` to your relations

db/models.py
```diff
@@ -46,3 +54,76 @@ class MovieSession(models.Model):
...
+        ]
+
+    def __str__(self) -> str:
+        return (
+            f"{self.movie_session.movie.title} {self.movie_session.show_time}"
+            f" (row: {self.row}, seat: {self.seat})"
+        )
+
+    def clean(self) -> None:
+        if not (1 <= self.row <= self.movie_session.cinema_hall.rows):
```
wQuelS, line 101: maybe it would be more readable if you create additional variable for `self.movie_session.cinema_hall` and then use `.rows` or `.seats_in_row` where needed

services/order.py
```diff
@@ -0,0 +1,26 @@
...
+        ticket["movie_session_id"] = ticket["movie_session"]
+        del ticket["movie_session"]
+        Ticket.objects.create(**ticket, order=new_order)
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
+    user_orders = Order.objects.all()
+    if username:
+        user = get_user_model().objects.get(username=username)
+        return user_orders.filter(user=user)
```
wQuelS, line 25: do you need this `return` ? You can just apply filters if `username` is provided and use one `return` outside the condition

wQuelS (changes requested):

services/order.py
```diff
@@ -0,0 +1,26 @@
...
+        ticket["movie_session_id"] = ticket["movie_session"]
+        del ticket["movie_session"]
+        Ticket.objects.create(**ticket, order=new_order)
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
+    user_orders = Order.objects.all()
+    if username:
+        user = get_user_model().objects.get(username=username)
+        user_orders = user_orders.filter(user=user)
```
YuriiKindrat, line 25: no need to get user model here, u can use lookups instead 
also don't think that user_orders is a good varialbe name for order.objects.all

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -41,8 +50,88 @@ def __str__(self) -> str:
...
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+    )
```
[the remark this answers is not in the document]
ZabFTFT: I set related name in Meta class.Is it wrong ?=)

PR #241, opened 2023-03-14
Solution

DavidGlusch:


viktoria-rybenchuk (approved):

db/models.py
```diff
@@ -46,3 +54,69 @@ class MovieSession(models.Model):
...
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(to=User, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{str(self.created_at)}"
```
wQuelS, line 68: you don't need f-string, since you are using `str()`

db/models.py
```diff
@@ -46,3 +54,69 @@ class MovieSession(models.Model):
...
+
+    class Meta:
+        ordering = ["-created_at"]
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        to=MovieSession, on_delete=models.CASCADE
```
wQuelS, line 76: add `related_name` to your relations

db/models.py
```diff
@@ -46,3 +54,69 @@ class MovieSession(models.Model):
...
+    )
+    order = models.ForeignKey(to=Order, on_delete=models.CASCADE)
```
wQuelS, line 78: you should refer to `AUTH_USER_MODEL`

db/models.py
```diff
@@ -46,3 +54,69 @@ class MovieSession(models.Model):
...
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(to=User, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{str(self.created_at)}"
+
+    class Meta:
+        ordering = ["-created_at"]
```
wQuelS, line 71: class Meta should be placed before methods. Fix everywhere

db/models.py
```diff
@@ -46,3 +54,69 @@ class MovieSession(models.Model):
...
+    seat = models.IntegerField()
+
+    def __str__(self) -> str:
+        return (f"{self.movie_session.movie.title} "
+                f"{self.movie_session.show_time} "
+                f"(row: {self.row}, seat: {self.seat})")
+
+    def clean(self) -> None:
+        cinema_hall_rows = self.movie_session.cinema_hall.rows
+        cinema_hall_seats = self.movie_session.cinema_hall.seats_in_row
```
wQuelS, line 89: maybe it is better to create just one `cinema_hall` variable and then use `.rows` or `.seats_in_row` when needed

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+
+        if date:
+            order.created_at = date
+
+        order.save()
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
```
wQuelS, line 26: use dictionary method `get()` as it is more reliable

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
+    return Order.objects.all()
```
wQuelS, line 36: maybe you should try to make a queryset with all orders and than apply filters if `username` is provided. With such approach you can avoid using double return

wQuelS (changes requested):

YuriiKindrat (approved):

PR #244, opened 2023-03-14
Solution

viktoria-rybenchuk:


Maytreyaya (approved):

db/models.py
```diff
@@ -46,3 +54,70 @@ class MovieSession(models.Model):
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(settings.AUTH_USER_MODEL,
+                             on_delete=models.CASCADE,
+                             default=None)
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+    class Meta:
+        ordering = ["-created_at"]
```
YuriiKindrat, line 69: class Meta should be placed before methods

services/movie_session.py
```diff
@@ -42,3 +46,16 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(
+        movie_session_id: int
+) -> list:
+    ticket = Ticket.objects.filter(
```
YuriiKindrat, line 54: ticket is bad name for variable here, does filter returns single ticket ?

YuriiKindrat (changes requested):

oksana-feshchenko (approved):

oksana-feshchenko (approved):
only not sure what save supposed to return

services/movie_session.py
```diff
@@ -42,3 +46,16 @@ def update_movie_session(
...
+        movie_session_id: int
+) -> list:
+    tickets = Ticket.objects.filter(
+        movie_session_id=movie_session_id
+    )
+    taken_seats = [
+        {"row": info.row, "seat": info.seat}
+        for info in tickets
+    ]
+    return taken_seats
```
YuriiKindrat, line 61: info is a bad name for variable, maybe it is better to use ticket? 
also u can use values method to simplify ur code

services/user.py
```diff
@@ -0,0 +1,48 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = get_user_model().objects.get(id=user_id)
```
YuriiKindrat, line 37: u can use get_user func here)

YuriiKindrat (changes requested):

services/movie_session.py
```diff
@@ -42,3 +46,14 @@ def update_movie_session(
...
+
+def get_taken_seats(
+        movie_session_id: int
+) -> list:
+    taken_seats = [
+        taken_seat for taken_seat in Ticket.objects.filter(
+            movie_session_id=movie_session_id
+        ).values("row", "seat")
+    ]
+    return taken_seats
```
YuriiKindrat, line 59: no need to create variable here

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #245, opened 2023-03-14
Movie Session Update Solution

Maytreyaya:


services/movie.py
```diff
@@ -1,3 +1,6 @@
+from __future__ import annotations
```
amber-marichi, line 1: im not sure you need this import here

services/movie_session.py
```diff
@@ -42,3 +42,9 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
+    tickets = (Ticket.objects.filter(movie_session=movie_session_id)
+               .values_list("row", "seat"))
+    return [{"row": ticket[0], "seat": ticket[1]} for ticket in tickets]
```
amber-marichi, line 50: why dont you just return <query>.values("row", "seat") instead of adding comprehension?

services/user.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import annotations
```
amber-marichi, line 1: again, i think you can remove this import

amber-marichi (changes requested):
nice done, few minor suggestions

services/movie.py
Maytreyaya: dont know why, but list[int] type in get movies doent work without it, so probably i will leave this one here.

amber-marichi (approved):
gj!

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+@transaction.atomic
+def create_order(tickets: list,
+                 username: str,
+                 date: datetime = None) -> Order:
+    user = get_user_model().objects.get(username=username)
+    order = Order.objects.create(user=user)
+
+    for ticket in tickets:
+        movie_session = MovieSession.objects.get(id=ticket["movie_session"])
+        Ticket.objects.create(movie_session=movie_session,
```
YuriiKindrat, line 19: Use movie_session_id argument and pass the id directly.

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #247, opened 2023-03-14
solution

amber-marichi:


kuznetsb (approved):
Nice work!

services/movie.py
```diff
@@ -6,15 +7,19 @@
...
-    if genres_ids:
+    if genres_ids is not None:
         queryset = queryset.filter(genres__id__in=genres_ids)
 
-    if actors_ids:
+    if actors_ids is not None:
         queryset = queryset.filter(actors__id__in=actors_ids)
 
+    if title is not None:
+        queryset = queryset.filter(title__contains=title)
```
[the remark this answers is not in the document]
amber-marichi: idk, probably just by habit. ill amend it, ty!

db/models.py
```diff
@@ -41,8 +49,67 @@ def __str__(self) -> str:
...
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+    class Meta:
+        ordering = ["-created_at"]
```
YuriiKindrat, line 71: class Meta should be placed before methods according to the best practicies
also why do u need to use f-string here?

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+        if date:
+            new_order.created_at = date
+            new_order.save()
+
+        for ticket_info in tickets:
+            Ticket.objects.create(
+                order=new_order,
+                movie_session=MovieSession.objects.get(
+                    id=ticket_info["movie_session"]
+                ),
```
YuriiKindrat, line 22: Use movie_session_id argument and pass the id directly.

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -41,8 +49,67 @@ def __str__(self) -> str:
...
+            })
+        if not (0 < self.seat <= cinema_seats):
+            raise ValidationError({
+                "seat": (
+                    f"seat number must be in available range: "
+                    f"(1, seats_in_row): (1, {cinema_seats})"
+                )
+            })
+
+    def save(self, *args, **kwargs) -> Ticket:
```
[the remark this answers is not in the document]
amber-marichi: nope, not really sure...

db/models.py
```diff
@@ -41,8 +50,75 @@ def __str__(self) -> str:
...
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
```
wQuelS, line 85: you've missed something I guess..

db/models.py
```diff
@@ -41,8 +50,75 @@ def __str__(self) -> str:
...
+            raise ValidationError({
+                "seat": (
+                    f"seat number must be in available range: "
+                    f"(1, seats_in_row): (1, {cinema_seats})"
+                )
+            })
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        return super(Ticket, self).save(*args, **kwargs)
```
wQuelS, line 124: you don't actually need `Ticket, self` here

wQuelS (changes requested):

db/models.py
amber-marichi: i hope you mean "related_name" because i cannot see if its something else, sorry :(

PR #251, opened 2023-03-15
Solution

MikhailLyvak:


db/models.py
```diff
@@ -17,7 +21,7 @@ def __str__(self) -> str:
 
 
 class Movie(models.Model):
-    title = models.CharField(max_length=255)
+    title = models.CharField(max_length=255, db_index=True)
```
NEkropo1, line 24: I guess this should be according to best practices in video, not as an argument to CharField.

(class Meta)

db/models.py
```diff
@@ -46,3 +50,73 @@ class MovieSession(models.Model):
...
+        Order,
+        on_delete=models.CASCADE
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=["row", "seat", "movie_session"],
+                                    name="unique_row_seat_session")
```
NEkropo1, line 84: Strange name for a ticket, more like we got unique_session ) But whatever

db/models.py
```diff
@@ -46,3 +50,73 @@ class MovieSession(models.Model):
...
+    def clean(self) -> None:
+        if not 1 < self.row <= self.movie_session.cinema_hall.rows:
+            raise ValidationError({
+                "row": [
+                    "row number must be in available range: "
+                    "(1, rows): "
+                    f"(1, {self.movie_session.cinema_hall.rows})"
+                ]
+            })
+        if not 1 < self.seat <= self.movie_session.cinema_hall.seats_in_row:
```
NEkropo1, line 101: I know it passes test, but i"m pretty sure, that seat with number 1 could be in there

NEkropo1 (changes requested):

db/models.py
MikhailLyvak: Here agree)

db/models.py
MikhailLyvak: But..... Create Meta class for 1 index it is creating of useless code. Maby in real project some space for future, but not here.

NEkropo1: Idk, I think it's better to always use best practices.

db/models.py
```diff
@@ -46,3 +50,73 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
```
YuriiKindrat, line 57: add related names everywhere

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
```
YuriiKindrat, line 28: what about get user model method here?

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -17,10 +21,16 @@ def __str__(self) -> str:
...
-    actors = models.ManyToManyField(to=Actor)
-    genres = models.ManyToManyField(to=Genre)
+    actors = models.ManyToManyField(
+        to=Actor,
+        related_name="actors"
+    )
+    genres = models.ManyToManyField(
+        to=Genre,
+        related_name="genres"
+    )
```
YuriiKindrat, line 33: incorrect names for related names, fix everywhere

YuriiKindrat (changes requested):

PR #255, opened 2023-03-15
solution

NEkropo1:


Danil228 (approved):

services/movie_session.py
```diff
@@ -42,3 +44,16 @@ def update_movie_session(
...
+def get_taken_seats(movie_session_id: int) -> List[Dict[str, int]]:
+    tickets = (
+        Ticket.objects
+        .filter(movie_session_id=movie_session_id)
+        .values("row", "seat")
+    )
+    taken_seats = [
+        {"row": ticket["row"], "seat": ticket["seat"]}
+        for ticket in tickets
+    ]
```
YuriiKindrat, line 58: why do u need this list comprehension if u are using values method ?

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+            row=ticket["row"],
+            seat=ticket["seat"],
+            movie_session_id=ticket["movie_session"],
+            order=order
+        )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        return User.objects.get(username=username).orders.all()
```
YuriiKindrat, line 33: use get_user_model method to get User

services/user.py
```diff
@@ -0,0 +1,31 @@
+from db.models import User
+
+
+def create_user(username: str,
+                password: str,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> None:
+    user = User.objects.create_user(username=username, password=password)
```
YuriiKindrat, line 9: same

services/user.py
```diff
@@ -0,0 +1,31 @@
...
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
```
YuriiKindrat, line 20: same

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -46,3 +53,79 @@ class MovieSession(models.Model):
...
+        using: object = None,
+        update_fields: object = None,
+    ) -> None:
+        self.full_clean()
+        return super().save(force_insert, force_update, using, update_fields)
+
+    def __str__(self) -> str:
+        show_time_str = self.movie_session.show_time.strftime(
+            "%Y-%m-%d %H:%M:%S"
+        )
```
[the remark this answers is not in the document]
NEkropo1: it could leave microseconds, while according to readme we need to show time only till seconds,

services/movie_session.py
NEkropo1: Because the task was to return list of dictionaries, while values method will return QuerySet of dictionaries.

services/user.py
NEkropo1: Btw, maybe better idea is to push declaration of this method on the start of the list and use implemented method on the remainings, since we already invented custom get_user with get_user_model() for every custom user?

services/movie_session.py
YuriiKindrat: u can just make list(QuerySet) to get list instead of queryset, no need to make loop for this

YuriiKindrat (changes requested):

services/movie_session.py
NEkropo1: Idk why I was sure, that, even though QS returns dicts with k-v, since we using values, it will return objects if I use list on QS) Thanks for clarifying.

YuriiKindrat (approved):

PR #257, opened 2023-03-15
Solution

oksanaaam:


myroslavatrschk (approved):

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+from django.db import transaction
+from db.models import Order, Ticket
+
+
+@transaction.atomic
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> None:
```
YuriiKindrat, line 13: return None?

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+            movie_session_id=ticket["movie_session"],
+            order=order,
+            row=ticket["row"],
+            seat=ticket["seat"]
+        )
+
+    return order
+
+
+def get_orders(username: str = None) -> None:
```
YuriiKindrat, line 32: same

YuriiKindrat (changes requested):

services/movie_session.py
```diff
@@ -24,6 +24,11 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
     return MovieSession.objects.get(id=movie_session_id)
 
 
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return list(MovieSession.objects.get(id=movie_session_id).
```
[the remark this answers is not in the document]
YuriiKindrat: also, it is unnecessary to use all here)

services/order.py
```diff
@@ -29,7 +29,7 @@ def create_order(
     return order
 
 
-def get_orders(username: str = None) -> None:
+def get_orders(username: str = None) -> Order:
```
YuriiKindrat, line 32: are u sure that filter\all methods returns Order instance ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #258, opened 2023-03-15
Solution

oksana-feshchenko:


ykopatko (approved):
gj

db/models.py
```diff
@@ -41,8 +48,90 @@ def __str__(self) -> str:
...
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+    class Meta:
+        ordering = ["-created_at"]
```
YuriiKindrat, line 74: class Meta should be placed before methods

db/models.py
```diff
@@ -41,8 +48,90 @@ def __str__(self) -> str:
...
+            f"{self.movie_session.show_time} "
+            f"(row: {self.row}, seat: {self.seat})"
+        )
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["movie_session", "row", "seat"], name="unique_placing"
+            )
+        ]
```
YuriiKindrat, line 103: same)

services/user.py
```diff
@@ -0,0 +1,46 @@
...
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+
+
+def get_user(user_id: int) -> get_user_model:
```
YuriiKindrat, line 23: u can use User for type annotation

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+            order.created_at = date
+            order.save()
+        [
+            order.tickets.create(
+                row=ticket["row"],
+                seat=ticket["seat"],
+                movie_session_id=ticket["movie_session"],
+            )
+            for ticket in tickets
+        ]
```
YuriiKindrat, line 25: no need to create list here, it is better to use for loops for creating

YuriiKindrat (changes requested):

PR #263, opened 2023-03-15
Solution

oigod:


services/user.py
```diff
@@ -0,0 +1,24 @@
...
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
+
+
+def update_user(user_id: int, password: str = None, **kwargs) -> None:
+    user = get_user(user_id)
+    if password:
+        user.set_password(password)
+        user.save()
+    User.objects.filter(id=user_id).update(**kwargs)
```
ostapT, line 24: same

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+
+def create_order(
+        tickets: list,
+        username: str,
+        date: str = None
+) -> None:
+    with transaction.atomic():
+        new_order = Order.objects.create(
+            user=User.objects.get(username=username)
+        )
```
ostapT, line 15: due to checklist we should use get_user_model().objects "its the best practice" instead of User.objects...

services/user.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        **kwargs
+) -> User:
+    new_user = User.objects.create_user(username=username, **kwargs)
+    new_user.set_password(password)
+    new_user.save()
```
ostapT, line 11: same with User.objects

ostapT (changes requested):

db/models.py
```diff
@@ -46,3 +60,57 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
```
YuriiKindrat, line 67: add related names everywhere

services/user.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        **kwargs
+) -> User:
+    new_user = get_user_model().objects.create_user(
+        username=username, **kwargs)
+    new_user.set_password(password)
```
YuriiKindrat, line 13: no need to use set_password here

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -41,8 +57,86 @@ def __str__(self) -> str:
 
 class MovieSession(models.Model):
     show_time = models.DateTimeField()
-    cinema_hall = models.ForeignKey(to=CinemaHall, on_delete=models.CASCADE)
-    movie = models.ForeignKey(to=Movie, on_delete=models.CASCADE)
+    cinema_hall = models.ForeignKey(
+        to=CinemaHall,
+        on_delete=models.CASCADE,
+        related_name="cinema_hall"
```
YuriiKindrat, line 63: incorrect related_name, fix everywhere

db/models.py
```diff
@@ -41,8 +57,86 @@ def __str__(self) -> str:
...
+                             name="unique_ticket_fields")
+        ]
+
+    def __str__(self) -> str:
+        return (f"{self.movie_session.movie.title} "
+                f"{self.movie_session.show_time} "
+                f"(row: {self.row}, seat: {self.seat})")
+
+    def clean(self) -> None:
+        if not (1 <= self.seat <= self.movie_session.cinema_hall.seats_in_row):
```
YuriiKindrat, line 116: u can create var for similar part and use it later `self.movie_session.cinema_hall`

PR #264, opened 2023-03-15
Solution

NazarHladaniuk:


skmiwa1 (approved):

db/models.py
```diff
@@ -46,3 +48,62 @@ class MovieSession(models.Model):
...
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
```
YuriiKindrat, line 59: you should refer to AUTH_USER_MODEL

db/models.py
```diff
@@ -46,3 +48,62 @@ class MovieSession(models.Model):
...
+
+    class Meta:
+        ordering = ["-user"]
```
YuriiKindrat, line 62: image
are u sure about ordering by user ?

db/models.py
```diff
@@ -46,3 +48,62 @@ class MovieSession(models.Model):
...
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        to=MovieSession, on_delete=models.CASCADE
+    )
+    order = models.ForeignKey(to=Order, on_delete=models.CASCADE)
```
YuriiKindrat, line 72: it is better to use related name for relations, fix everywhere

db/models.py
```diff
@@ -46,3 +48,62 @@ class MovieSession(models.Model):
...
+        constraints = [
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="row_seat_movie_session"
+            )
+        ]
+
+    def clean(self) -> None:
+        c_hall_rows = self.movie_session.cinema_hall.rows
+        c_hall_seat = self.movie_session.cinema_hall.seats_in_row
```
YuriiKindrat, line 86: u can just create variable for self.movie_session.cinema_hall and than call rows/seats etc. no need to create variables for attributes

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+            seat=ticket_data["seat"],
+            movie_session_id=ticket_data["movie_session"],
+            order=order,
+        )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
+    else:
```
YuriiKindrat, line 34: unnecessary else here

services/user.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> User:
+    user = User.objects.create_user(username=username, password=password)
```
YuriiKindrat, line 13: don't use User model, use get_user_model method instead, fix everywhere

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #269, opened 2023-03-15
Solution

myroslavatrschk:


db/models.py
```diff
@@ -46,3 +55,75 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
```
YuriiKindrat, line 62: add related names everywhere

db/models.py
```diff
@@ -46,3 +55,75 @@ class MovieSession(models.Model):
...
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket"
+            )
+        ]
+
+    def clean(self) -> None:
+        rows = self.movie_session.cinema_hall.rows
+        seats = self.movie_session.cinema_hall.seats_in_row
```
YuriiKindrat, line 96: no need to create vars for attributes, u can create only one var for similar part

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = get_user_model().objects.get(id=user_id)
```
YuriiKindrat, line 38: u can use get_user func here

YuriiKindrat (changes requested):

PR #274, opened 2023-03-16
Solution

VladyslavCherkashyn:


db/models.py
```diff
@@ -46,3 +55,64 @@ class MovieSession(models.Model):
...
+
+    def save(
+            self, force_insert: bool = False,
+            force_update: bool = False,
+            using: Any = None,
+            update_fields: Any = None
+    ) -> "Ticket":
+        self.full_clean()
+
+        return super(Ticket, self).save()
```
YuriiKindrat, line 115: what about args ? this method takes some args 107-112 lines

services/movie_session.py
```diff
@@ -42,3 +41,11 @@ def update_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    taken_seats = Ticket.objects.filter(movie_session_id=movie_session_id)
+    return [
+        {"row": ticket.row, "seat": ticket.seat}
+        for ticket in taken_seats
+    ]
```
YuriiKindrat, line 51: try to use values method to simplify ur code

db/models.py
YuriiKindrat: u didn't pass args to a super.save() method

YuriiKindrat (changes requested):

Y-Havryliv (approved):
Nice!

PR #276, opened 2023-03-16
Solution

Litvinenko23:


db/models.py
```diff
@@ -41,8 +50,88 @@ def __str__(self) -> str:
...
+    )
     movie = models.ForeignKey(to=Movie, on_delete=models.CASCADE)
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
```
YuriiKindrat, line 64: add related names everywhere

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def create_order(
+    tickets: list[Ticket], username: str, date: str = None
+) -> Order:
+    with transaction.atomic():
+        try:
+            user = get_user_model().objects.get(username=username)
+        except get_user_model().DoesNotExist:
+            user = get_user_model().objects.create(username=username)
```
YuriiKindrat, line 15: will u create user without password ?

services/user.py
```diff
@@ -0,0 +1,41 @@
...
+
+def update_user(
+    user_id: int,
+    username: str = None,
+    password: str = None,
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None,
+) -> User:
+    user = get_user_model().objects.get(id=user_id)
```
YuriiKindrat, line 27: u can use get user func here

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -41,8 +50,97 @@ def __str__(self) -> str:
...
-    cinema_hall = models.ForeignKey(to=CinemaHall, on_delete=models.CASCADE)
+    cinema_hall = models.ForeignKey(
+        to=CinemaHall,
+        related_name="movie_sessions",
+        on_delete=models.CASCADE,
+    )
     movie = models.ForeignKey(to=Movie, on_delete=models.CASCADE)
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
```
YuriiKindrat, line 61: same here

YuriiKindrat (changes requested):

PR #278, opened 2023-03-16
Solution

tytomyr:


db/models.py
```diff
@@ -46,3 +57,59 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(settings.AUTH_USER_MODEL,
```
YuriiKindrat, line 64: add related names everywhere

db/models.py
```diff
@@ -46,3 +57,59 @@ class MovieSession(models.Model):
...
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.movie_session.movie.title}" \
+               f" {self.movie_session.show_time}" \
+               f" (row: {self.row}, seat: {self.seat})"
```
YuriiKindrat, line 83: use brackets instead of linebreakers

db/models.py
```diff
@@ -46,3 +57,59 @@ class MovieSession(models.Model):
...
+                            " (1, rows): (1, 10)"]})
+
+        if not (1 <= self.seat <= self.movie_session.cinema_hall.seats_in_row):
+            raise ValidationError(
+                {
+                    "seat": ["seat number must be in available range:"
+                             " (1, seats_in_row): (1, 12)"]})
+
+    class Meta:
+        unique_together = ["row", "seat", "movie_session"]
```
YuriiKindrat, line 99: class Meta should be placed before methods

services/movie_session.py
```diff
@@ -42,3 +42,15 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    session = MovieSession.objects.get(id=movie_session_id)
```
YuriiKindrat, line 48: no need to get session here, use lookups in Ticket query

YuriiKindrat: also try to use values method to simplify ur code

services/order.py
```diff
@@ -0,0 +1,30 @@
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order, User, Ticket
+
+
+@transaction.atomic
+def create_order(tickets: list, username: str, date: str = None) -> Order:
+    order = Order.objects.create(user=User.objects.get(username=username))
```
YuriiKindrat, line 9: what about get user model method here ?

services/order.py
```diff
@@ -0,0 +1,30 @@
...
+            row=ticket["row"],
+            seat=ticket["seat"],
+            order=order
+        )
+
+    return order
+
+
+def get_orders(username: str = None) -> QuerySet:
+    with transaction.atomic():
```
YuriiKindrat, line 27: why do u use transaction here ?

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    return user.save()
+
+
+def get_user(user_id: int) -> get_user_model:
```
YuriiKindrat, line 23: same

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(user_id: int,
+                username: str = None,
+                password: (str | int) = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> get_user_model:
```
YuriiKindrat, line 32: same

YuriiKindrat (changes requested):

db/models.py
YuriiKindrat: didn't fix that

db/models.py
YuriiKindrat: same

services/movie_session.py
YuriiKindrat: didn't fix that

services/order.py
YuriiKindrat: same

services/order.py
YuriiKindrat: same

services/user.py
```diff
@@ -0,0 +1,49 @@
+from django.contrib.auth import get_user_model
+
+
+def create_user(username: str,
+                password: (str | int),
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> get_user_model:
```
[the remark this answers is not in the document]
YuriiKindrat: didn't fix that

YuriiKindrat (changes requested):

Y-Havryliv (approved):
Well done!

PR #279, opened 2023-03-16
'Solution'

Zoriana-Dvulit:


db/models.py
```diff
@@ -46,3 +61,34 @@ class MovieSession(models.Model):
...
+        super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=["movie_session", "row", "seat"], name="unique_ticket")
+        ]
+
+
+class User(AbstractUser):
+    pass
```
ArsenPidhoretskyi, line 94: Add new line at the end of file

services/movie.py
```diff
@@ -1,11 +1,12 @@
 from django.db.models import QuerySet
-
+from django.db import transaction
 from db.models import Movie
 
 
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
ArsenPidhoretskyi, line 9: it accepts Optional[str] because default parameter is None

ArsenPidhoretskyi: Fix in other places(only your code)

ArsenPidhoretskyi (changes requested):
Fix flake8

Bohdan-Salamakha (changes requested):
fix tests
`ModuleNotFoundError: No module named 'django_extensions'
`update ur requirements.txt

db/models.py
```diff
@@ -46,3 +65,48 @@ class MovieSession(models.Model):
...
+        on_delete=models.CASCADE
+    )
+    order = models.ForeignKey(to=Order, on_delete=models.CASCADE)
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.movie_session.movie.title}" \
+               f" {self.movie_session.show_time}" \
+               f" (row: {self.row}, seat: {self.seat})"
```
Bohdan-Salamakha, line 82:
```
return (f"{self.movie_session.movie.title}"
        f" {self.movie_session.show_time}"
        f" (row: {self.row}, seat: {self.seat})")

```

services/movie.py
Bohdan-Salamakha: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

services/movie_session.py
```diff
@@ -25,10 +26,10 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
...
 
 def update_movie_session(
-    session_id: int,
-    show_time: str = None,
-    movie_id: int = None,
-    cinema_hall_id: int = None,
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None,
```
Bohdan-Salamakha, line 32: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket, User
+
+
+@transaction.atomic
+def create_order(
+        tickets: list,
```
Bohdan-Salamakha, line 12: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

services/user.py
```diff
@@ -0,0 +1,37 @@
...
+    return user
+
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
```
Bohdan-Salamakha, line 22: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):
v1

services/movie.py
ArsenPidhoretskyi: @Zoriana-Dvulit not yet fixed

services/order.py
ArsenPidhoretskyi: @Zoriana-Dvulit not filex

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+) -> Order:
+
+    user = User.objects.get(username=username)
+    if date is not None:
+        date = datetime.datetime.strptime(date, "%Y-%m-%d %H:%M")
+        order = Order.objects.create(user=user)
+        order.created_at = date
+        order.save()
+    else:
+        order = Order.objects.create(user=user)
```
ArsenPidhoretskyi, line 24: No need to duplicate code do as:
```suggestion
        date = datetime.datetime.strptime(date, "%Y-%m-%d %H:%M") if date is not None else None
        order = Order.objects.create(user=user, created_at=date)
```
Maybe like this?

ArsenPidhoretskyi (changes requested):

services/order.py
Zoriana-Dvulit: I fixed all, except last one, because an error raises.

Zoriana-Dvulit: @ArsenPidhoretskyi  і @Bohdan-Salamakha , є надія, що сьогодні буду пити каву за завершення свого 9 модуля з цією задачею -:)

db/models.py
```diff
@@ -46,3 +65,48 @@ class MovieSession(models.Model):
...
+    order = models.ForeignKey(to=Order, on_delete=models.CASCADE)
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    def __str__(self) -> str:
+        return (f"{self.movie_session.movie.title}"
+                f" {self.movie_session.show_time}"
+                f" (row: {self.row}, seat: {self.seat})")
+
+    def clean(self) -> NoReturn:
```
olena-hul, line 84: Interesting annotation) But you could have used None instead (no need to fix it now, just remember for future)

services/order.py
olena-hul: @Zoriana-Dvulit It will be better if this is fixed. Let's talk in slack about this

db/models.py
Zoriana-Dvulit: I fix it, thank you very much!

Bohdan-Salamakha (approved):

PR #280, opened 2023-03-16
Solution

StreetNik:


db/models.py
```diff
@@ -46,3 +58,56 @@ class MovieSession(models.Model):
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.CASCADE
+    )
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
ArsenPidhoretskyi, line 73: Use string format instead of just f-string

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
ArsenPidhoretskyi, line 10: It accepts Optional[str] because None as default
Fix in other places(Just your code)

ArsenPidhoretskyi (changes requested):

PR #281, opened 2023-03-17
'Solution'

Bohdan-Kovalenko1992:


db/models.py
```diff
@@ -46,3 +54,68 @@ class MovieSession(models.Model):
...
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
```
YuriiKindrat, line 65: you should refer to AUTH_USER_MODEL, also add related names everywhere

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
...
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return [{
+        "row": ticket["ticket__row"],
+        "seat": ticket["ticket__seat"]
+    } for ticket in MovieSession.objects.filter(
+        id=movie_session_id
+    ).values("ticket__row", "ticket__seat")]
```
YuriiKindrat, line 53: try to use Ticket model to simplify ur code

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+from db.models import Order, Ticket, User
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> Order:
+    with transaction.atomic():
+        order = Order.objects.create(user=User.objects.get(username=username))
```
YuriiKindrat, line 15: what about get user model method ?

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+            )
+
+        return order
+
+
+def get_orders(username: str = None) -> QuerySet:
+    orders = Order.objects.all()
+
+    if username is not None:
+        orders = orders.filter(user=User.objects.get(username=username))
```
YuriiKindrat, line 37: no need to get user here, use lookups instead

PR #282, opened 2023-03-17
'Solution'

natalie-goriela:


services/movie.py
```diff
@@ -1,11 +1,15 @@
...
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+    genres_ids: Optional[list] = None,
+    actors_ids: Optional[list] = None,
```
Bohdan-Salamakha, line 11: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

services/movie.py
```diff
@@ -25,16 +32,17 @@ def get_movie_by_id(movie_id: int) -> Movie:
 def create_movie(
     movie_title: str,
     movie_description: str,
-    genres_ids: list = None,
-    actors_ids: list = None,
+    genres_ids: Optional[list] = None,
+    actors_ids: Optional[list] = None,
```
Bohdan-Salamakha, line 36: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

services/movie.py
```diff
@@ -25,16 +32,17 @@ def get_movie_by_id(movie_id: int) -> Movie:
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
Bohdan-Salamakha, line 38: use @transaction.atomic() decorator

Bohdan-Salamakha (changes requested):
v1

ArsenPidhoretskyi (approved):
Well done!

PR #283, opened 2023-03-18
Solution

oksana0410:


db/models.py
```diff
@@ -46,3 +51,76 @@ class MovieSession(models.Model):
...
+    class Meta:
+        constraints = [UniqueConstraint(
+            fields=["row", "seat", "movie_session"],
+            name="unique_row_seat_movie_session"
+        )]
+
+    def __str__(self) -> str:
+        return f"{self.movie_session.movie.title} "\
+               f"{self.movie_session.show_time} "\
+               f"(row: {self.row}, seat: {self.seat})"
```
YuriiKindrat, line 98: use brackets instead of linebreakers

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #288, opened 2023-03-25
Solution

arsenmakovei:


services/user.py
```diff
@@ -0,0 +1,53 @@
...
+    if username:
+        user.username = username
+    if password:
+        user.set_password(password)
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
```
Vasyl-Smutok, line 51: `DRY` 

here and 18-25 lines

Vasyl-Smutok (changes requested):
Also you can use `black` to fix indentation

Vasyl-Smutok (approved):
Great 👍

PR #289, opened 2023-03-26
Solution

Bohdan2001007:


services/movie.py
```diff
@@ -1,11 +1,16 @@
...
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+    genres_ids: List[int] = None,
+    actors_ids: List[int] = None,
```
ArsenPidhoretskyi, line 12: Optional[List[int]]

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+            seat=ticket_data["seat"]
+        )
+
+    return order
+
+
+def get_orders(username: Optional[str] = None) -> QuerySet:
+    if username:
+        user = User.objects.get(username=username)
+        return Order.objects.filter(user=user)
```
ArsenPidhoretskyi, line 35: Simplify this. You can filter by related field

services/user.py
```diff
@@ -0,0 +1,50 @@
...
+
+User = get_user_model()
+
+
+def create_user(
+    username: str,
+    password: str,
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None
```
ArsenPidhoretskyi, line 12: Optional[str]

ArsenPidhoretskyi (changes requested):

services/movie.py
```diff
@@ -23,18 +29,19 @@ def get_movie_by_id(movie_id: int) -> Movie:
...
 
 def create_movie(
-    movie_title: str,
-    movie_description: str,
-    genres_ids: list = None,
-    actors_ids: list = None,
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list = None,
+        actors_ids: list = None,
```
Bohdan-Salamakha, line 35: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #290, opened 2023-03-26
Solution

Rename925:


services/user.py
```diff
@@ -0,0 +1,61 @@
...
+    if username:
+        user.username = username
+    if password:
+        user.set_password(password)
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if username:
+        user.last_name = last_name
```
Vasyl-Smutok, line 59: `DRY`

here and 21-30 lines

db/models.py
```diff
@@ -46,3 +55,65 @@ class MovieSession(models.Model):
...
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey("db.User", on_delete=models.CASCADE)
```
Vasyl-Smutok, line 66: add related names everywhere

services/user.py
```diff
@@ -0,0 +1,61 @@
...
+def update_user(
+        user_id: int,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        email: Optional[str] = None,
+        first_name: Optional[str] = None,
+        last_name: Optional[str] = None
+) -> None:
+
+    user = get_user_model().objects.get(id=user_id)
```
Vasyl-Smutok, line 48: use `get_user` here

Vasyl-Smutok (changes requested):

services/user.py
Rename925: How can i change it?

services/movie.py
```diff
@@ -15,13 +19,17 @@ def get_movies(
     if actors_ids:
         queryset = queryset.filter(actors__id__in=actors_ids)
 
+    if title:
```
Grayder0152, line 22:
```suggestion
    if title is not None:
```

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+        date: Optional[str] = None
+) -> None:
+
+    user = User.objects.get(username=username)
+
+    new_order = Order.objects.create(
+        user=user
+    )
+
+    if date:
```
Grayder0152, line 23:
```suggestion
    if date is not None:

```

services/user.py
Grayder0152: Create a function and use it in this case and in line 19-28

Grayder0152 (changes requested):

ArsenPidhoretskyi (changes requested):

services/user.py
ArsenPidhoretskyi: @Rename925 not fixed

ArsenPidhoretskyi (changes requested):

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+            seat=ticket["seat"],
+            order=new_order
+        )
+
+
+def get_orders(username: Optional[str] = None) -> QuerySet:
+    query_set = Order.objects.all()
+    if username:
+        user = get_user_model().objects.get(username=username)
+        query_set = query_set.filter(user_id=user.id)
```
ArsenPidhoretskyi, line 40: What you know about filtering by related field?
https://stackoverflow.com/questions/17132450/django-filter-on-field-of-related-model
This method should have only one query

ArsenPidhoretskyi (changes requested):

PR #291, opened 2023-03-27
Solution

YehorKorn:


services/movie.py
```diff
@@ -1,14 +1,19 @@
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
+    title: str = None,
```
Grayder0152, line 8: add Optional type

services/movie.py
```diff
@@ -1,14 +1,19 @@
...
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
 ) -> QuerySet:
     queryset = Movie.objects.all()
 
+    if title:
```
Grayder0152, line 14:
```suggestion
    if title is not None:
```

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
+        movie = Movie.objects.create(
+            title=movie_title,
+            description=movie_description,
+        )
+        if genres_ids:
```
Grayder0152, line 41: the same as above

Grayder0152 (changes requested):

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+            ticket["movie_session_id"] = ticket["movie_session"]
+            del ticket["movie_session"]
+
+            Ticket.objects.create(
+                order=current_order,
+                **ticket
+            )
+
+
+def get_orders(username: Optional = None) -> QuerySet:
```
Grayder0152, line 35:
```suggestion
def get_orders(username: Optional[str] = None) -> QuerySet:

```

Grayder0152 (changes requested):

PR #293, opened 2023-03-29
Solution1

serhii-skivka:


services/movie.py
```diff
@@ -28,13 +35,13 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```
Bohdan-Salamakha, line 38: use @transaction.atomic() decorator

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+
+from db.models import Order, User, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: Optional[datetime.datetime] = None
+) -> None:
+    with transaction.atomic():
```
Bohdan-Salamakha, line 15: use @transaction.atomic() decorator

Bohdan-Salamakha (changes requested):
v1

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+        tickets: list[dict],
+        username: str,
+        date: Optional[datetime.datetime] = None
+) -> None:
+    user = User.objects.get(username=username)
+    order = Order.objects.create(user=user)
+
+    if date:
+        order.created_at = date
+        order.save()
```
EdAlekseiev, line 21: This code executes two SQL queries. Please, upgrade this code so that it executes one SQL query.

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+    for ticket in tickets:
+        Ticket.objects.create(
+            order=order,
+            movie_session_id=ticket["movie_session"],
+            row=ticket["row"],
+            seat=ticket["seat"]
+        )
+
+
+def get_orders(username: Optional[str] = None) -> Order | QuerySet[Order]:
```
EdAlekseiev, line 32: Typehint of return values is incorrect. Order.objects.filter returns QuerySet too.

EdAlekseiev (changes requested):

services/order.py
serhii-skivka: if i change the code for example: order = Order.objects.create(user=user, created_at=date if date else None)   it gonna execute 1 sql query but it will result in AssertionError 
AssertionError: assert datetime.datetime(2023, 3, 30, 12, 49, 23, 153446) == datetime.datetime(2020, 11, 10, 14, 40)

Bohdan-Salamakha (approved):

PR #295, opened 2023-04-02
Solution

lotrik84:


services/user.py
```diff
@@ -0,0 +1,57 @@
...
+    if email:
+        user.email = email
+
+    if first_name:
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
```
Vasyl-Smutok, line 57: `DRY` 

you repeat yourself here and 16-25 lines

services/user.py
```diff
@@ -0,0 +1,57 @@
...
+    return User.objects.get(id=user_id)
+
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
```
Vasyl-Smutok, line 38: use Optional for annotation

Vasyl-Smutok (changes requested):

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
ArsenPidhoretskyi, line 10: Correct type is Optional[str] because it by default accepts None
Fix in other places(Just your code)

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+        if date:
+            order.created_at = date
+            order.save()
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
+
+    return Order.objects.all()
```
ArsenPidhoretskyi, line 35:
```suggestion
    queryset = Order.objects.all()
    if username:
        queryset = queryset.filter(user__username=username)

    return queryset
```

ArsenPidhoretskyi: This one is better. Because when we will add new parameters it will be more hardly to change this logic

services/user.py
```diff
@@ -0,0 +1,43 @@
...
+        if key == "email":
+            user.email = value
+        if key == "first_name":
+            user.first_name = value
+        if key == "last_name":
+            user.last_name = value
+        if key == "username":
+            user.username = value
+        if key == "password":
+            user.set_password(value)
```
ArsenPidhoretskyi, line 15: Can you explain what changed? As for me it will be the same but without loop. With this modification, code is more unreadable
If you want to unify better to do as:
```
for field, value in arguments.items():
    if value is not None:
        if field in ["email", "first_name", "last_name", "username"]:
            setattr(user, field, value)
        elif field == "password":
            user.set_password(value)
```

ArsenPidhoretskyi (changes requested):

services/movie.py
```diff
@@ -15,6 +19,9 @@ def get_movies(
     if actors_ids:
         queryset = queryset.filter(actors__id__in=actors_ids)
 
+    if title:
```
Grayder0152, line 22:
```suggestion
    if title is not None:
```

services/movie_session.py
```diff
@@ -42,3 +42,10 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return [
+        {"row": ticket.row, "seat": ticket.seat}
+        for ticket in Ticket.objects.filter(movie_session_id=movie_session_id)
+    ]
```
Grayder0152, line 51: use `values` method

Grayder0152 (changes requested):

Grayder0152 (approved):

PR #296, opened 2023-04-03
'Solution'

Mazurinka29:


services/movie.py
```diff
@@ -1,11 +1,15 @@
+from typing import List
+import init_django_orm  # noqa: F401
```
ArsenPidhoretskyi, line 2: why you need this import here?

services/movie.py
```diff
@@ -1,11 +1,15 @@
...
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+    genres_ids: List[int] = None,
+    actors_ids: List[int] = None,
+    title: str = None,
```
ArsenPidhoretskyi, line 12: It accepts Optional[str] and Optional[List[int]]
Because as default they accepts None
Fix in all places

services/order.py
```diff
@@ -0,0 +1,40 @@
+from typing import List
+
+import init_django_orm  # noqa: F401 щоб імпортувати,
```
ArsenPidhoretskyi, line 3: You does not need this import here

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+            )
+
+    return new_ticket
+
+
+def get_orders(username: str = None) -> QuerySet:
+    queryset = Order.objects.all()
+    if username:
+        find_user = get_user_model().objects.get(username=username)
+        queryset = queryset.filter(user_id=find_user.id)
```
ArsenPidhoretskyi, line 38: You can filter by related field:
user__username

services/order.py
ArsenPidhoretskyi: Fix in all places

services/user.py
```diff
@@ -0,0 +1,64 @@
...
+        updated_user.username = username
+
+    if password:
+        updated_user.set_password(password)
+
+    if email:
+        updated_user.email = email
+
+    if first_name:
+        # updated_user.first_names = first_name
```
ArsenPidhoretskyi, line 59: remove this code

ArsenPidhoretskyi (changes requested):

PR #302, opened 2023-04-05
Develop

U-Ra-All:


services/movie.py
```diff
@@ -1,11 +1,16 @@
...
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+    genres_ids: Optional[list[int]] = None,
+    actors_ids: Optional[list[int]] = None,
+    title: str = None
```
Bohdan-Salamakha, line 13: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie.py
```diff
@@ -15,26 +20,32 @@ def get_movies(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```
Bohdan-Salamakha, line 39: use @transaction.atomic() decorator

Bohdan-Salamakha (changes requested):
v1

services/movie.py
U-Ra-All: Done

services/movie.py
U-Ra-All: Done

olena-hul (approved):

olena-hul:
Good job!

U-Ra-All:
> Good job!

Thanks!

PR #304, opened 2023-04-07
Solution_v1

bohdan-yatsyna:


services/movie_session.py
```diff
@@ -42,3 +44,11 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
```
ArsenPidhoretskyi, line 49: list of what?

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+from django.contrib.auth import get_user_model
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket
+
+
+@transaction.atomic
+def create_order(
+        tickets: list,
```
ArsenPidhoretskyi, line 12: tickets list of what?

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+            order=new_order
+        )
+
+    return new_order
+
+
+def get_orders(username: Optional[str] = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
+    return Order.objects.all()
```
ArsenPidhoretskyi, line 37: Better to do like that, because when you add new params it will be simple to modify
```suggestion
    queryset = Order.objects.all()
    if username:
        queryset = queryset.filter(user__username=username)
    return queryset
```

services/user.py
```diff
@@ -0,0 +1,35 @@
...
+        username: str,
+        password: str,
+        **kwargs
+) -> None:
+    new_user = get_user_model().objects.create_user(
+        username=username,
+        password=password,
+    )
+
+    for key, value in kwargs.items():
```
ArsenPidhoretskyi, line 16: rename key to field_name

services/user.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def get_user(user_id: int) -> User:
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(user_id: int, **kwargs) -> None:
+    user = get_user_model().objects.get(id=user_id)
+
+    for key, value in kwargs.items():
```
ArsenPidhoretskyi, line 29: rename key to field_name

ArsenPidhoretskyi (changes requested):

services/order.py
bohdan-yatsyna: Thank you for all the remarks. Updated already.

ArsenPidhoretskyi (approved):
Well done!

PR #307, opened 2023-04-11
solution

MkrtychKhachatrian:


services/user.py
```diff
@@ -0,0 +1,54 @@
...
+    if username:
+        user.username = username
+    if password:
+        user.set_password(password)
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
```
ArsenPidhoretskyi, line 52: create a separate function and move this logic to. Duplicate code 19-24 lines

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+                movie_session_id=ticket["movie_session"]
+            )
+
+        return order
+
+
+def get_orders(username: Optional[str] = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
+    return Order.objects.all()
```
ArsenPidhoretskyi, line 36:
```suggestion
    queryset = Order.objects.all()
    if username:
        queryset = queryset.filter(user__username=username)
    return queryset
```

ArsenPidhoretskyi: In future it will be more easily to support code

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+from django.db.models import QuerySet
+
+from db.models import Ticket, Order
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: Optional[str] = None) -> Order:
+    with transaction.atomic():
```
vasylhnatiuk, line 14: use decorator @transaction.atomic instead of this

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #311, opened 2023-04-21
Solution

Roman28101:


ArsenPidhoretskyi (changes requested):

db/models.py
```diff
@@ -46,3 +66,50 @@ class MovieSession(models.Model):
...
+    movie_session = models.ForeignKey(to=MovieSession,
+                                      on_delete=models.CASCADE)
+    order = models.ForeignKey(to=Order, on_delete=models.CASCADE)
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.movie_session.movie.title} " \
+               f"{self.movie_session.show_time} " \
+               f"(row: {self.row}, seat: {self.seat})"
```
Grayder0152, line 81: Use `()` instead of `\`

db/models.py
```diff
@@ -46,3 +66,50 @@ class MovieSession(models.Model):
...
+        if not 1 <= self.seat <= seats:
+            raise ValidationError(
+                {"seat": [f"seat number must be in available range: "
+                          f"(1, seats_in_row): (1, {seats})"]}
+            )
+
+    def save(
+            self,
+            force_insert: Optional[bool] = False,
+            force_update: Optional[bool] = False,
```
Grayder0152, line 100: Why `Optional`?

services/user.py
```diff
@@ -0,0 +1,63 @@
...
+
+def set_data(
+        user: User,
+        username: Optional[str],
+        email: Optional[str] = None,
+        first_name: Optional[str] = None,
+        last_name: Optional[str] = None,
+        password: Optional[str] = None
+) -> None:
+    if username:
```
Grayder0152, line 16:
```suggestion
    if username is not None:
```

services/user.py
```diff
@@ -0,0 +1,63 @@
...
+        user.username = username
+    if email:
```
Grayder0152, line 18:
```suggestion
    if email is not None:
```

Grayder0152 (changes requested):

Grayder0152 (approved):

PR #316, opened 2023-04-30
Solution

olenaliuby:


services/movie_session.py
```diff
@@ -26,19 +30,30 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
...
         movie_session.cinema_hall_id = cinema_hall_id
     movie_session.save()
 
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    movie_session = MovieSession.objects.get(id=movie_session_id)
```
ArsenPidhoretskyi, line 52: Use prefetch_related

services/user.py
```diff
@@ -0,0 +1,58 @@
...
+        username=username,
+        password=password,
+    )
+
+    if email is not None:
+        user.email = email
+    if first_name is not None:
+        user.first_name = first_name
+    if last_name is not None:
+        user.last_name = last_name
```
ArsenPidhoretskyi, line 24: You can move this to a separate function. Because this code the same as in update_user

ArsenPidhoretskyi (changes requested):

services/movie_session.py
olenaliuby: Something like that?
`movie_session = MovieSession.objects.prefetch_related('tickets').get(id=movie_session_id)`

I thought, maybe to go with:
```
return list(
        Ticket.objects.filter(movie_session_id=movie_session_id)
        .values("row", "seat")
    )
```
It seems more clear and shorter. I didn't think about it the first time.

services/user.py
```diff
@@ -0,0 +1,75 @@
+from typing import Optional
+
+from django.contrib.auth import get_user_model
+
+from db.models import User
+
+
+def create_or_update_email_and_f_l_name(
```
Vasyl-Smutok, line 8: bad function name 

try something  like this 
`modify_user_data`, `update_user_information`

Vasyl-Smutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #324, opened 2023-05-07
Preliminary solution

oleg-potichnyi:


Bohdan-Salamakha (changes requested):
fix flake8

db/migrations/0001_initial.py
```diff
@@ -5,54 +5,102 @@
 
 
 class Migration(migrations.Migration):
-
     initial = True
 
-    dependencies = [
-    ]
+    dependencies = []
```
olena-hul, line 10: Fix the tests

olena-hul (changes requested):

PR #327, opened 2023-05-11
Solution

xtina999:


services/movie.py
```diff
@@ -1,19 +1,22 @@
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
+    title: str = None,
```
ArsenPidhoretskyi, line 8: It accepts Optional[str] because default parameter is None
Check in all places(only your code)

services/user.py
```diff
@@ -0,0 +1,52 @@
...
+        password=password,
+
+    )
+
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
```
ArsenPidhoretskyi, line 24: Create additional function and remove duplicate code(lines 46-51)

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #332, opened 2023-05-16
Develop

nickkozlov90:


services/movie.py
```diff
@@ -28,13 +35,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
EdAlekseiev, line 38: Use transaction.atomic() as decorator instead of context manager

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+                    id=ticket["movie_session"]
+                ),
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+
+def get_orders(
+    username: Optional[str] = None
+) -> Union[QuerySet, User]:
```
EdAlekseiev, line 35: It does not return `User`. It returns only queryset of `Order`

services/user.py
```diff
@@ -0,0 +1,54 @@
...
+        user_id: int,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        email: Optional[str] = None,
+        first_name: Optional[str] = None,
+        last_name: Optional[str] = None,
+) -> None:
+    user = User.objects.get(id=user_id)
+
+    for arg in locals().keys():
```
EdAlekseiev, line 43: What is this cycle for?

EdAlekseiev (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #336, opened 2023-05-23
Solution

Phoenix-Erazer:


services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
ArsenPidhoretskyi, line 10: It accepts Optional[str] because None as default. Fix in all places(just your code)

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+    user = get_user_model().objects.create_user(
+        username=username,
+        password=password,
+    )
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
```
ArsenPidhoretskyi, line 22: Move this code to separate function and call it here. Because now duplicate code with lines 44-49

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Good job!

PR #339, opened 2023-05-28
Develop

VladyslavOladko:


services/user.py
```diff
@@ -0,0 +1,48 @@
...
+    new_user = get_user_model().objects.create_user(
+        username=username,
+        password=password,
+    )
+    if email:
+        new_user.email = email
+    if first_name:
+        new_user.first_name = first_name
+    if last_name:
+        new_user.last_name = last_name
```
vsmutok, line 21: DRY 

here 42 -48 lines

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
vsmutok, line 36: Use transaction.atomic() as decorator instead of context manager

vsmutok (changes requested):

services/movie.py
```diff
@@ -6,9 +7,13 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
ArsenPidhoretskyi, line 10: It accepts Optional[str] because None as default. Fix in all places(only your code)

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    taken_seats = []
+    movie_session = MovieSession.objects.get(id=movie_session_id)
+    tickets = Ticket.objects.filter(movie_session=movie_session)
```
ArsenPidhoretskyi, line 50: You can filter by related field no need to do 2 queries

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #342, opened 2023-05-31
Solution

VTeteruk:


services/movie.py
```diff
@@ -1,13 +1,15 @@
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
+    title: str = None,
```
ArsenPidhoretskyi, line 8: It accepts Optional[str] because None as default

services/movie_session.py
```diff
@@ -42,3 +42,10 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    movie_session = MovieSession.objects.get(id=movie_session_id)
```
ArsenPidhoretskyi, line 48: You can filter by Ticket:
Ticket.objects.filter(movie_sesssion_id=movie_session_id)
Also you can use .values()

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+            order=new_order,
+            row=ticket.get("row"),
+            seat=ticket.get("seat")
+        )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
+    return Order.objects.all()
```
ArsenPidhoretskyi, line 34: Better to do like this:
```suggestion
    queryset = Order.objects.all()
    if username:
        queryset = queryset.filter(user__username=username)
    return queryset
```

ArsenPidhoretskyi: In future it will be more easy to support code

services/movie.py
ArsenPidhoretskyi: Fix in all places(only your code)

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+    new_user = get_user_model().objects.create_user(
+        username=username, password=password
+    )
+
+    if email:
+        new_user.email = email
+    if first_name:
+        new_user.first_name = first_name
+    if last_name:
+        new_user.last_name = last_name
```
ArsenPidhoretskyi, line 23: You can move this code to separate function and remove duplicate of code from create_user and update_user

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+
+    new_user.save()
+    return new_user
+
+
+def get_user(user_id: int) -> QuerySet:
```
ArsenPidhoretskyi, line 29: It returns User, not queryset

ArsenPidhoretskyi (changes requested):

services/movie.py
```diff
@@ -1,13 +1,17 @@
...
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+    title: Optional[str] = None,
+    genres_ids: Optional[list[int]] = None,
+    actors_ids: Optional[list[int]] = None,
 ) -> QuerySet:
-    queryset = Movie.objects.all()
+    queryset = Movie.objects.filter(title__icontains=title)
```
ArsenPidhoretskyi, line 14: But if title is None?
You must check:
if title:

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #345, opened 2023-06-02
Solution

Niki-Alex:


services/order.py
```diff
@@ -0,0 +1,37 @@
...
+                seat=ticket["seat"]
+            )
+        return order
+
+
+def get_orders(username: Optional[str] = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(
+            user__username=username
+        ).values_list("created_at")
```
ArsenPidhoretskyi, line 36: Why you return only created_at?
You must return order objects

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+    user_create = User.objects.create_user(
+        username=username,
+        password=password,
+    )
+    if email:
+        user_create.email = email
+    if first_name:
+        user_create.first_name = first_name
+    if last_name:
+        user_create.last_name = last_name
```
ArsenPidhoretskyi, line 23: Move this code to separate function and remove duplicate of code from update_user

ArsenPidhoretskyi (changes requested):

PR #348, opened 2023-06-05
Solution

Dimdimior:


services/movie.py
```diff
@@ -28,13 +35,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket, MovieSession
+
+
+def create_order(tickets: list[dict],
+                 username: str,
+                 date: Optional[str] = None) -> None:
+    with transaction.atomic():
```
EdAlekseiev, line 13: The same issue as above

services/user.py
```diff
@@ -0,0 +1,44 @@
...
+from django.contrib.auth import get_user_model
+
+from db.models import User
+
+
+def create_user(username: str,
+                password: str,
+                email: Optional[str] = "",
+                first_name: Optional[str] = "",
+                last_name: Optional[str] = "") -> None:
```
EdAlekseiev, line 12: You have to use Optional only if your variable can be None. Your variables can't be None, so there are not optional

db/models.py
```diff
@@ -46,3 +56,74 @@ class MovieSession(models.Model):
...
+                            f"(1, seats_in_row): "
+                            f"(1, {seats})"
+                }
+            )
+
+    def save(self,
+             force_insert: bool = False,
+             force_update: bool = False,
+             using: str = None,
+             update_fields: list[str] = None) -> bool:
```
EdAlekseiev, line 123: These arguments can be optional, improve type hints, please.

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #352, opened 2023-06-08
'Solution'

IhorPokr:


services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
Bohdan-Salamakha, line 10: use typehint like: Optional[YOUR_VALUE_TYPE] = None
for example: list that contains string elements and can be None has typehint Optional[list[str]]

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+
+from django.db import transaction
+from django.contrib.auth import get_user_model
+from django.db.models import QuerySet
+
+from db.models import Ticket, Order
+
+
+@transaction.atomic
+def create_order(tickets: list[dict], username: str, date: str = None) -> None:
```
Bohdan-Salamakha, line 11: use typehint like: Optional[YOUR_VALUE_TYPE] = None
for example: list that contains string elements and can be None has typehint Optional[list[str]]

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+    for ticket in tickets:
+        Ticket.objects.create(
+            movie_session_id=ticket["movie_session"],
+            order=order,
+            row=ticket["row"],
+            seat=ticket["seat"]
+        )
+
+
+def get_orders(username: str = None) -> QuerySet:
```
Bohdan-Salamakha, line 36: use typehint like: Optional[YOUR_VALUE_TYPE] = None
for example: list that contains string elements and can be None has typehint Optional[list[str]]

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #355, opened 2023-06-13
Added `Tickets and orders` functional to MovieTheatre project.

N1khto:


db/models.py
```diff
@@ -46,3 +56,66 @@ class MovieSession(models.Model):
...
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE
+    )
+
+    def __str__(self) -> str:
+        return str(self.created_at)
```
vasylhnatiuk, line 69:
```suggestion
        return f"{self.created_at.strftime('%Y-%m-%d %H:%M:%S')}"
```

services/user.py
```diff
@@ -0,0 +1,59 @@
...
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = get_user(user_id=user_id)
+
+    if username:
```
vasylhnatiuk, line 44: try to avoid duplicating code, use loop and setattr

vasylhnatiuk (changes requested):

PR #357, opened 2023-06-15
Solution

OksanaAdamchuk:


services/order.py
```diff
@@ -0,0 +1,38 @@
...
+                movie_session=movie_session,
+                order=new_order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+            )
+
+        return new_order
+
+
+def get_orders(username: str = None) -> QuerySet:
```
ArsenPidhoretskyi, line 34: It accepts Optiona[str] because None by default. Fix in all places with all types

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #362, opened 2023-06-22
Solution

OlehOryshchuk:


db/models.py
```diff
@@ -46,3 +55,75 @@ class MovieSession(models.Model):
...
+        to=settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
olena-hul, line 72: No need to use f-strings, if there is only 1 parameter inside. Use str(self.created_at)

db/models.py
```diff
@@ -46,3 +55,75 @@ class MovieSession(models.Model):
...
+                         f"(1, {seats_in_row})"]
+            })
+
+    def save(
+            self,
+            force_insert: bool = False,
+            force_update: bool = False,
+            using: str = None,
+            update_fields: list = None
+    ) -> None:
```
olena-hul, line 123: Add Optional to annotations where the default argument can be None

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+            new_ticket.clean()
+            users_tickets.append(new_ticket)
+
+        Ticket.objects.bulk_create(users_tickets)
+
+
+def get_orders(username: Optional[str] = None) -> QuerySet:
+    return (users.objects.get(username=username).orders.all()
+            if username
+            else Order.objects.all())
```
olena-hul, line 41: Fix formatting

Better to use Order query and filter by user__username here

olena-hul (changes requested):

services/movie.py
```diff
@@ -28,13 +34,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
Bohdan-Salamakha, line 37: use @transaction.atomic() decorator

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+
+users = get_user_model()
+
+
+def create_order(
+        tickets: list[Ticket],
+        username: str,
+        date: Optional[str] = None
+) -> None:
+    with transaction.atomic():
```
Bohdan-Salamakha, line 16: use @transaction.atomic() decorator

Bohdan-Salamakha (changes requested):

db/models.py
```diff
@@ -46,3 +56,75 @@ class MovieSession(models.Model):
...
+        to=settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return str(self.created_at)
```
EdAlekseiev, line 73: Use `strftime` method for convert `datetime` to `str`

EdAlekseiev (changes requested):

Bohdan-Salamakha (approved):

PR #365, opened 2023-06-27
Develop

Andrew13X:


db/migrations/0001_initial.py
```diff
@@ -1,58 +1,118 @@
...
+
+    initial = True
+
+    dependencies = [
+        ('auth', '0012_alter_user_first_name_max_length'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='User',
```
vasylhnatiuk, line 21: change ' to ", use Black

db/models.py
```diff
@@ -41,8 +54,81 @@ def __str__(self) -> str:
...
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.movie_session.movie.title}" \
```
vasylhnatiuk, line 91: use () instead of \

vasylhnatiuk (changes requested):

db/models.py
Andrew13X: Got it. Thanks

db/migrations/0001_initial.py
Andrew13X: Could you please. I have to use "Black" instead of "to" in 0001_initial.py?

vasylhnatiuk: change ' to ",  'User' -> "User". do it here and in all of files. Black does it automatically

vasylhnatiuk: https://pypi.org/project/black/

vasylhnatiuk (changes requested):

db/migrations/0001_initial.py
Andrew13X: Thanks! It really helped

ArsenPidhoretskyi (approved):
Well done!

PR #367, opened 2023-07-04
Solution

gleblitvinenko:


services/movie.py
```diff
@@ -6,9 +7,13 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
ArsenPidhoretskyi, line 10: It accepts Optional[str] because None as default. Fix in all places(just your code)

services/movie_session.py
```diff
@@ -42,3 +42,9 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    movie_session = get_movie_session_by_id(movie_session_id)
+    return [{"row": ticket.row, "seat": ticket.seat}
+            for ticket in movie_session.ticket_set.all()]
```
ArsenPidhoretskyi, line 50: You do two queries when can do only one

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+                row=ticket_data["row"],
+                seat=ticket_data["seat"])
+
+        order.save()
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        user = User.objects.get(username=username)
+        return Order.objects.filter(user=user)
```
ArsenPidhoretskyi, line 32: You can filter by related name. No need to get a user. Read how to filter by related fields

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+            Ticket.objects.create(
+                order_id=order.id,
+                movie_session_id=ticket_data["movie_session"],
+                row=ticket_data["row"],
+                seat=ticket_data["seat"])
+
+        order.save()
+
+
+def get_orders(username: Optional[str] = None) -> QuerySet:
```
vasylhnatiuk, line 31:
```suggestion
def get_orders(username: Optional[str] = None) -> QuerySet[Order]:

```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #369, opened 2023-07-04
Solution

Kirontiko:


services/order.py
```diff
@@ -0,0 +1,35 @@
...
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+        order.save()
+
+    return order
+
+
+def get_orders(username: str = None) -> Order | QuerySet:
```
AntonShpakovych, line 30: QuerySet[Order]

services/user.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def create_user(username: str,
+                password: str,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> User:
+    user = get_user_model().objects.create_user(username=username,
+                                                password=password
+                                                )
```
AntonShpakovych, line 13: If u change optional default parametrs to  `""` u can write 
```return get_user_model().objects.create_user(
        username=username,
        password=password,
        email=email,
        first_name=first_name,
        last_name=last_name
    )```. Basically if u don't put smth in email etc django fill with all values with `""`. I am not sure mb its bad practise

**Review by AntonShpakovych — APPROVED** [review]

File: db/models.py [hunk 3]
```diff
@@ -46,3 +54,67 @@ class MovieSession(models.Model):
...
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
```
YuriiKindrat, line 65: it is better to use AUTH_USER_MODEL from settings here

db/models.py
```diff
@@ -46,3 +54,67 @@ class MovieSession(models.Model):
...
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+    class Meta:
+        ordering = ["-created_at"]
```
YuriiKindrat, line 71: methods usually writes after class Meta

services/movie_session.py
```diff
@@ -42,3 +42,8 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return list(Ticket.objects.filter(
+        movie_session_id=movie_session_id).values("row", "seat"))
```
YuriiKindrat, line 49: bad formatting here

services/user.py
YuriiKindrat: it is better to use None as defaul value here

services/user.py
```diff
@@ -0,0 +1,48 @@
...
+                password: str,
+                email: str = "",
+                first_name: str = "",
+                last_name: str = "") -> User:
+    user = get_user_model().objects.create_user(username=username,
+                                                password=password,
+                                                email=email,
+                                                first_name=first_name,
+                                                last_name=last_name
+                                                )
```
YuriiKindrat, line 16: will user be created correctly with email as None etc. it seems to me it is better to make check here )

YuriiKindrat (changes requested):

services/movie_session.py
Kirontiko: I have only one idea how to fix formatting here, I hope that I`m not wrong

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+def get_user(user_id: int) -> User:
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> None:
```
Oleg-s-s, line 33: Yurii mentioned:
_will user be created correctly with email as None_
and you made None as a default value) This is wrong, f.e. email field cannot be None, fix everywhere

Oleg-s-s (changes requested):

services/user.py
Kirontiko: But if we don't apply a value to a charfield in django - it'll set blank by default or i'm wrong?

Oleg-s-s: I see, you talked about validation on line 36-49, it still can be improved by taking kwargs instead of checking each value 1 by 1, and use for loop

Oleg-s-s: Same in create user

Oleg-s-s (changes requested):

services/user.py
Kirontiko: Oh, now I understand, I`ll fix it soon, thank you!

Oleg-s-s (approved):
:+1:

PR #370, opened 2023-07-04
Solution

AntonShpakovych:


services/order.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime.date = None
+) -> Order:
+    with transaction.atomic():
+        order = Order.objects.create(
+            user=get_user_model().objects.get(username=username)
```
Kirontiko, line 17: I think it's not best practice not to create variable for query, but idk, tbh

Kirontiko (approved):
I

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+            order.save()
+
+        for ticket in tickets:
+            row, seat, movie_session_id = ticket.values()
+            Ticket.objects.create(
+                order=order,
+                row=row,
+                seat=seat,
+                movie_session_id=movie_session_id
+            )
```
Oleg-s-s, line 31: You can also try to use bulk_create here to not overload the db. Not necessary, just my opinion.

AntonShpakovych: Важливо враховувати деякі обмеження bulk_create(). Зокрема, при використанні цього методу не викликаються методи save()

AntonShpakovych: So in this example probably we can't use bulk_create because we have validation which connected to save(). And I had error in test like: ticket with this movie_session already presented

Oleg-s-s: Very strange... Maybe you did some mistake cause it should work in both ways

Oleg-s-s (approved):
Good job

PR #371, opened 2023-07-05
Solution

DmytroTsehelnyk:


db/models.py
```diff
@@ -46,3 +55,81 @@ class MovieSession(models.Model):
...
+        related_name="tickets"
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket_row_seat_movie_session"
```
volodymyr-komarnyckyi, line 93: I'm not sure but as for me it's not the best name

services/movie.py
```diff
@@ -6,9 +7,13 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
 ) -> QuerySet:
     queryset = Movie.objects.all()
 
+    if title:
+        queryset = queryset.filter(title__contains=title)
```
volodymyr-komarnyckyi, line 15: I'd recommend using icontains here for a broader search

volodymyr-komarnyckyi (approved):
Well done!

services/user.py
```diff
@@ -0,0 +1,47 @@
...
+
+from db.models import User
+
+
+def create_user(
+    username: str,
+    password: str,
+    email: str = "",
+    first_name: str = "",
+    last_name: str = ""
```
YuriiKindrat, line 11: it is better to use None here, also u can use kwargs here)

YuriiKindrat (changes requested):

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+from db.models import Order, Ticket
+
+
+def create_order(
+    tickets: list[dict],
+    username: str,
+    date: datetime = None
+) -> Order:
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
```
YuriiKindrat, line 16: no need to create additional variable for user here

YuriiKindrat (approved):

PR #372, opened 2023-07-05
'Solution'

ant-komarov:


diana-shyrokikh (approved):

db/models.py
```diff
@@ -46,3 +58,67 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE)
```
YuriiKindrat, line 67: bad formmating here(

db/models.py
```diff
@@ -46,3 +58,67 @@ class MovieSession(models.Model):
...
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return self.created_at.strftime("%Y-%m-%d %H:%M:%S")
```
YuriiKindrat, line 73: no need to use strftime here, u can use str instead

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #377, opened 2023-07-06
Solution #1

erikagayan:


db/models.py
```diff
@@ -46,3 +55,71 @@ class MovieSession(models.Model):
...
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket_row_seat_movie_session"
+            )
+        ]
+
+    def clean(self) -> None:
+        if not self.row <= self.movie_session.cinema_hall.rows:
```
YevheniiIovchu, line 87: This check will be ok if you used PositiveIntegerField. You can receive self.row -3, for example, and it will be no error.

db/models.py
```diff
@@ -46,3 +55,71 @@ class MovieSession(models.Model):
...
+            raise ValidationError(
+                {
+                    "row": [
+                        "row number must be in available range: (1, rows): "
+                        f"(1, {self.movie_session.cinema_hall.rows})"
+                    ]
+                }
+            )
+
+        if not self.seat <= self.movie_session.cinema_hall.seats_in_row:
```
YevheniiIovchu, line 97: And the same here

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+            order.save()
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
+                order_id=order.id,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+        order.save()
```
YevheniiIovchu, line 26: You made save() in if check. If you don't change anything after creation no need to use save() more.

services/user.py
```diff
@@ -0,0 +1,48 @@
...
+        first_name: str = "",
+        last_name: str = ""
+) -> User:
+    return get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        email=email,
+        first_name=first_name,
+        last_name=last_name
+    )
```
YevheniiIovchu, line 18: You need to create user with optional fields (email, first_name, last_name) just in that case if they were provided.

YevheniiIovchu (changes requested):

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+            order.created_at = date
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
+                order_id=order.id,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+        order.save()
```
YevheniiIovchu, line 25: Maybe I explained not the best way. I meant you had already save() in if check, so you don't need additional one after the remaining part of your code. You need to make save() just in if check like below:
        if date:
            order.created_at = date
            order.save()

services/user.py
```diff
@@ -5,17 +5,25 @@
 def create_user(
         username: str,
         password: str,
-        email: str = "",
-        first_name: str = "",
-        last_name: str = ""
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
```
YevheniiIovchu, line 10: You can replace emal, first_name, last_name with **kwargs. Then you need to create some list/tuple with "email", "first_name", "last_name", and create your object 'User' by username and password. Then you need to unpack your kwargs and iterate one by one, checking if such key exist in your kwargs.items() and checking if value of this key is not "" or None. In this case after your check you can use setattr() to add this value to your 'User' object

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+        user.set_password(password)
+
+    if email:
+        user.email = email
+
+    if first_name:
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
```
YevheniiIovchu, line 54: And make the same here. update_user(user_id: int, password: str, **kwargs)

YevheniiIovchu (changes requested):

services/movie_session.py
```diff
@@ -42,3 +42,8 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return list(Ticket.objects.filter(
+        movie_session_id=movie_session_id).values("row", "seat"))
```
YuriiKindrat, line 49: bad formmating here

services/order.py
```diff
@@ -0,0 +1,36 @@
+from django.contrib.auth import get_user_model
+from db.models import Order, Ticket
+from django.db import transaction
+from django.db.models import QuerySet
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None) -> Order:
```
YuriiKindrat, line 10: same

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    if email:
+        user_data["email"] = email
+
+    if first_name:
+        user_data["first_name"] = first_name
+
+    if last_name:
+        user_data["last_name"] = last_name
```
YuriiKindrat, line 24: try to optimize this

YuriiKindrat (changes requested):

PR #378, opened 2023-07-06
done

maccarets:


services/movie_session.py
```diff
@@ -42,3 +42,10 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
+    movie_session = MovieSession.objects.get(id=movie_session_id)
+
+    return [{"row": ticket.row, "seat": ticket.seat}
+            for ticket in movie_session.ticket_set.all()]
```
YuriiKindrat, line 51: could u use values method here?

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+from django.db import transaction
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
+    with transaction.atomic():
+        user = User.objects.get(username=username)
```
YuriiKindrat, line 12: what about get_user_model method ?

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = User.objects.get(id=user_id)
```
YuriiKindrat, line 38: what about get_user_model method here? 
also u can use get_user func implemented higher

YuriiKindrat (changes requested):

Oleg-s-s (approved):
:+1:

PR #379, opened 2023-07-06
My solution

kovaliskoveronika:


services/order.py
```diff
@@ -0,0 +1,38 @@
...
+        username: str,
+        date: str = None
+) -> None:
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
+        new_order = Order.objects.create(user_id=user.id)
+        if date:
+            new_order.created_at = date
+
+        new_order.save()
```
Altahoma, line 19: `save()` will be calls twice (for the first time during creation of `new_order`) no matter if date provided, so it's better to move `save()` into `if` body.

services/user.py
```diff
@@ -0,0 +1,49 @@
+from db.models import User
+from django.contrib.auth import get_user_model
```
Altahoma, line 2: It is worth importing according to PEP8 requirements.

Altahoma (changes requested):

services/user.py
kovaliskoveronika: I'm not sure that I got your point about it. So can you look at this part again in the new commit?)

services/user.py
```diff
@@ -1,6 +1,7 @@
-from db.models import User
 from django.contrib.auth import get_user_model
 
+from db.models import User
+
```
Altahoma: Great job! I meant [imports order](https://peps.python.org/pep-0008/#imports).

Altahoma (approved):

Oleg-s-s (approved):
:+1:

PR #380, opened 2023-07-06
solution

Gvilion:


services/movie_session.py
```diff
@@ -42,3 +42,10 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: id) -> list[dict]:
+    return [
+        {"row": ticket.row, "seat": ticket.seat} for ticket
+        in Ticket.objects.filter(movie_session__id=movie_session_id)
+    ]
```
YuriiKindrat, line 51: try to use values method instead

YuriiKindrat (changes requested):
all tests should pass

Oleg-s-s (approved):
:+1:

PR #381, opened 2023-07-06
solution

oleksiikolii:


services/movie.py
```diff
@@ -15,13 +17,17 @@ def get_movies(
     if actors_ids:
         queryset = queryset.filter(actors__id__in=actors_ids)
 
+    if title:
+        queryset = queryset.filter(title__icontains=title)
+
     return queryset
```
Andriy-Sydorenko, line 24: You can set more descriptive name instead of `queryset` here

Andriy-Sydorenko (changes requested):

db/models.py
```diff
@@ -46,3 +56,60 @@ class MovieSession(models.Model):
...
+                        f" (1, {self.movie_session.cinema_hall.seats_in_row})"
+            })
+
+    def save(self, force_insert: bool = False,
+             force_update: bool = False,
+             using: bool = None,
+             update_fields: bool = None) -> callable:
+        self.full_clean()
+        return super(Ticket, self).save(force_insert, force_update, using,
+                                        update_fields)
```
YuriiKindrat, line 111: it is better to place one arg per line, for better readability

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+
+@transaction.atomic()
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
+    user = get_user_model().objects.get(username=username)
+
+    order = Order.objects.create(user=user)
```
YuriiKindrat, line 16: it seems to me, that there is no need to create additional variable )
u can place ur get_user_model method inside an Order)
Also, u are using save method later, so u can use just Order for creating without objects manager

services/user.py
```diff
@@ -0,0 +1,45 @@
+from django.contrib.auth import get_user_model
+
+from db.models import User
+
+
+def create_user(username: str, password: str,
+                **kwargs
+                ) -> None:
```
YuriiKindrat, line 8: bad formmating here(

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #382, opened 2023-07-06
Solution

volodymyr-komarnyckyi:


YKonoplyov (approved):
Well done👍

db/models.py
```diff
@@ -46,3 +55,66 @@ class MovieSession(models.Model):
...
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
```
YuriiKindrat, line 66: it is better to use AUTH_USER_MODEL from settings instead of User

services/user.py
```diff
@@ -0,0 +1,48 @@
...
+
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = "",
+        first_name: str = "",
+        last_name: str = ""
```
YuriiKindrat, line 11: it is better to use None instead, also try to use kwargs here)

services/user.py
```diff
@@ -0,0 +1,48 @@
...
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None) -> None:
```
YuriiKindrat, line 34: bad formmating(

YuriiKindrat (changes requested):

services/user.py
```diff
@@ -0,0 +1,40 @@
...
+        **kwargs: dict[str]
+) -> User:
+    user = get_user_model().objects.create_user(
+        username=username,
+        password=password
+    )
+
+    for el in kwargs:
+        if el in ["username", "email", "first_name", "last_name"]:
+            setattr(user, el, kwargs[el])
```
YuriiKindrat, line 18: what is el ? it seems to me that el is bad name for var

services/user.py
```diff
@@ -0,0 +1,40 @@
...
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(
+        user_id: int,
+        **kwargs: dict[str]
+) -> None:
+    user_update = get_user(user_id)
+
+    for el in kwargs:
```
YuriiKindrat, line 34: same (

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #383, opened 2023-07-06
Solution

Andriy-Sydorenko:


db/models.py
```diff
@@ -46,3 +54,60 @@ class MovieSession(models.Model):
...
+                "seat": f"seat number must be in available range: "
+                        f"(1, seats_in_row): "
+                        f"(1, {self.movie_session.cinema_hall.seats_in_row})"
+            })
+
+    def save(self,
+             force_insert: bool = False,
+             force_update: bool = False,
+             using: str = None,
+             update_fields: Iterable[str] = None) -> None:
```
YevheniiIovchu, line 108: I guess type(update_fields) just list[str], but no matter

services/user.py
```diff
@@ -0,0 +1,44 @@
...
+
+from db.models import User
+
+
+def create_user(username: str,
+                password: str,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> None:
+    user = User.objects.create_user(username=username, password=password)
```
YevheniiIovchu, line 11: Don't forget to use get_user_model insead of User. 3rd point of code style in checklist

services/user.py
```diff
@@ -0,0 +1,44 @@
...
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
```
YevheniiIovchu, line 17: You can make just 1 if check and put all optional params in **kwargs. Then using settatr you can assign new value to these params.

services/user.py
```diff
@@ -0,0 +1,44 @@
...
+    if username:
+        user.username = username
+    if password:
+        user.set_password(password)
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
```
YevheniiIovchu, line 43: And use here also **kwargs and setattr.

YevheniiIovchu (changes requested):

services/user.py
```diff
@@ -0,0 +1,31 @@
...
+def update_user(user_id: int,
+                password: str = None,
+                **kwargs) -> None:
+    user = get_user(user_id)
+
+    if password:
+        user.set_password(password)
+
+    for data_field, data in kwargs.items():
+        setattr(user, data_field, data)
```
YevheniiIovchu, line 29: fields = ["email", ...]
You need to make additional check 'if data and data_field in fields' to check that your data is not None or your data is not "" and to check that your function shouldn't try to create object if somebody write 'udpate_user(1, "qwertyui", my_child_dream='popularity')' because model User/AbstractUser has no column 'my_child_dream' and it causes User.DoesNotExist error

services/user.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def create_user(username: str,
+                password: str,
+                **kwargs) -> None:
+    get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        **kwargs
+    )
```
YevheniiIovchu, line 13: Make here like in example below, because you can get two error.
1) create_user("username", "password", email=None) because email=CharField(isnull=False by default)
2) create_user("username", "password", emaill="some@gmail.com") because no exist such field 'emaill' with 'll'

YevheniiIovchu (changes requested):

YevheniiIovchu (approved):

YuriiKindrat (approved):

PR #384, opened 2023-07-06
Solution

YKonoplyov:


db/models.py
```diff
@@ -46,3 +60,58 @@ class MovieSession(models.Model):
...
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=["row", "seat", "movie_session"],
+                                    name="unique_row_seat_movie_session")
```
volodymyr-komarnyckyi, line 86: In my opinion it's not the best name for variable, maybe unique_ticket or sth like this would be better

volodymyr-komarnyckyi (approved):

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+from django.db.models import QuerySet
+from datetime import datetime
+
+from db.models import Order, User, Ticket, MovieSession
+
+
+def create_order(tickets: list[dict], username: str, date: str = None) -> None:
+    with transaction.atomic():
+        order = Order.objects.create(
+            user=User.objects.get(username=username)
```
YuriiKindrat, line 11: use get user model method instead

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #385, opened 2023-07-06
Solution

Po-SvinSkipodla:


services/order.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@transaction.atomic
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> None:
+    new_order = Order.objects.create(
+        user_id=User.objects.get(username=username).id
```
oleksiikolii, line 16: according to checklist you need to use get_user_model() instead of User

oleksiikolii (changes requested):

db/models.py
```diff
@@ -46,3 +56,61 @@ class MovieSession(models.Model):
...
+        self.full_clean()
+        return super().save()
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket_row_seat_movie_session"
+            )
+        ]
```
YuriiKindrat, line 112: class meta should be placed before all ur methods

services/order.py
YuriiKindrat: agree

services/user.py
```diff
@@ -0,0 +1,54 @@
...
+
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = "",
+        last_name: str = ""
```
YuriiKindrat, line 11: use None instead of ""
also try to use kwargs here

services/user.py
```diff
@@ -0,0 +1,54 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    updated_user = get_user_model().objects.get(id=user_id)
```
YuriiKindrat, line 42: u can use method created higher here

YuriiKindrat (changes requested):

services/user.py
YuriiKindrat: u didn't fix that, "" is still here

services/user.py
```diff
@@ -0,0 +1,54 @@
...
+        first_name: str = "",
+        last_name: str = ""
+) -> None:
+    new_user = get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        email=email,
+        first_name=first_name,
+        last_name=last_name
+    )
```
YuriiKindrat, line 19: why don't u create user without email,first/last name here ?

YuriiKindrat (changes requested):

Oleg-s-s (approved):
:+1:

PR #386, opened 2023-07-06
'Solution'

yakhlamova:


kovaliskoveronika (approved):

db/models.py
```diff
@@ -46,3 +53,71 @@ class MovieSession(models.Model):
...
+                                        using,
+                                        update_fields)
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_row_seat_movie_session"
+            )
+        ]
```
YuriiKindrat, line 119: it is better to create class meta before all ur methods

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -17,7 +24,7 @@ def __str__(self) -> str:
 
 
 class Movie(models.Model):
-    title = models.CharField(max_length=255)
+    title = models.CharField(max_length=255, db_index=True)
     description = models.TextField()
     actors = models.ManyToManyField(to=Actor)
     genres = models.ManyToManyField(to=Genre)
```
Oleg-s-s, line 30: related names?

Oleg-s-s (approved):

PR #387, opened 2023-07-06
Implemented Order, Ticket, User models and services for them

Altahoma:


services/movie_session.py
```diff
@@ -42,3 +42,9 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    tickets = get_movie_session_by_id(movie_session_id).ticket_set.all()
+
+    return [{"row": ticket.row, "seat": ticket.seat} for ticket in tickets]
```
Tarasidze, line 50: Interesting decision, I like it

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+from django.db.models import QuerySet
+
+from db.models import User, Order, MovieSession, Ticket
+
+
+def create_order(
+    tickets: list[dict], username: str, date: datetime = None
+) -> Order:
+    with transaction.atomic():
+        user = User.objects.get(username=username)
```
Tarasidze, line 13: The transaction should be "Atomicy" - remove unnecessary operations from the context manager

db/models.py
```diff
@@ -46,3 +55,74 @@ class MovieSession(models.Model):
...
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=("row", "seat"), name="unique_row_seat"
```
Tarasidze, line 82: Maybe, there should be a unique combination: row, seat, and movie_session

Tarasidze (changes requested):

db/models.py
Altahoma: I agree, the uniqueness should be checked for the combination of these three attributes: `row`, `seat`, and `movie_session`.

Tarasidze (approved):

services/movie_session.py
YuriiKindrat: it seems to me, u could use `values` method here

services/order.py
YuriiKindrat: what about get_user_model method method ?

YuriiKindrat (changes requested):

services/order.py
Altahoma: Missed one, fixed.

db/models.py
```diff
@@ -22,6 +26,11 @@ class Movie(models.Model):
     actors = models.ManyToManyField(to=Actor)
     genres = models.ManyToManyField(to=Genre)
```
Oleg-s-s, line 27: related_name

Oleg-s-s (approved):

PR #389, opened 2023-07-07
Solutions 1.01

Tarasidze:


services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return list(MovieSession.objects.filter(id=movie_session_id).values(
+        row=F("ticket__row"),
+        seat=F("ticket__seat")))
+    # MovieSession.objects.get(id=movie_session_id).
+    # ticket.values("row", "seat")
```
mariiahorbova, line 52: remove unnecessary comments

db/models.py
```diff
@@ -41,8 +46,70 @@ def __str__(self) -> str:
...
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+
+class Ticket(models.Model):
+    row = models.PositiveIntegerField()
+    seat = models.PositiveIntegerField()
+    movie_session = models.ForeignKey(to="MovieSession",
+                                      on_delete=models.CASCADE,
+                                      related_name="ticket")
```
mariiahorbova, line 72: wrong indentation, you have the right one when creating new tickets:
Ticket.objects.create(
                row=ticket["row"],
                seat=ticket["seat"],
                movie_session=movie,
                order=order
            )
like this

db/models.py
```diff
@@ -41,8 +46,70 @@ def __str__(self) -> str:
...
+    order = models.ForeignKey(to="Order", on_delete=models.CASCADE)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"], name="unique seat")]
```
mariiahorbova, line 78: wrong indentation

db/models.py
```diff
@@ -41,8 +46,70 @@ def __str__(self) -> str:
...
+            force_insert: bool = False,
+            force_update: bool = False,
+            using: any = None,
+            update_fields: any = None
+    ) -> None:
+        self.full_clean()
+        super(Ticket, self).save(force_insert,
+                                 force_update,
+                                 using,
+                                 update_fields)
```
mariiahorbova, line 106: wrong indentation

services/order.py
```diff
@@ -0,0 +1,47 @@
...
+        MovieSession.objects.get(id=movie_session_id)
+    except MovieSession.DoesNotExist:
+        print(f"MovieSession with id: {movie_session_id}, does not exist")
+
+    return True
+
+
+def create_order(tickets: list[Ticket],
+                 username: str,
+                 date: str = None) -> None:
```
mariiahorbova, line 18: wrong indentation

mariiahorbova (changes requested):

services/order.py
Tarasidze: Thank you for your review)

db/models.py
```diff
@@ -19,8 +21,11 @@ def __str__(self) -> str:
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
-    actors = models.ManyToManyField(to=Actor)
-    genres = models.ManyToManyField(to=Genre)
+    actors = models.ManyToManyField(to="Actor")
+    genres = models.ManyToManyField(to="Genre")
```
YuriiKindrat, line 25: what about related names ?

services/order.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def create_order(
+        tickets: list[Ticket],
+        username: str,
+        date: str = None
+) -> None:
+
+    current_user = get_user_model().objects.get(username=username)
+    order = Order(user=current_user)
```
YuriiKindrat, line 23: there is no need to create additional variable here, u can make this directly in Order )

db/models.py
YuriiKindrat: also why do u need quotes here?)

YuriiKindrat (changes requested):

services/order.py
Tarasidze: In my opinion, this object provides less queries to the database. Also, when "data" is provided, we add it to our object below. And only then, in context manager, we save order to the database.

db/models.py
```diff
@@ -19,8 +21,11 @@ def __str__(self) -> str:
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
-    actors = models.ManyToManyField(to=Actor)
-    genres = models.ManyToManyField(to=Genre)
+    actors = models.ManyToManyField(to=Actor, related_name="movie")
+    genres = models.ManyToManyField(to=Genre, related_name="movie")
```
Oleg-s-s, line 25: according to checklist from previous task its better to use plural name for related_name

services/order.py
```diff
@@ -0,0 +1,51 @@
...
```
Oleg-s-s, line 25: number of queries remain the same, there is a use case for creating additional variable if you are referencing to it later, but here its just an additional line of code)

Oleg-s-s (changes requested):

services/order.py
Tarasidze: sorry I misunderstood, now it is ok, I think)

Oleg-s-s (approved):
:+1:

PR #390, opened 2023-07-07
Tickets and orders

IhorVoskoboinikov:


services/order.py
```diff
@@ -0,0 +1,40 @@
...
+        tickets: List[Dict[str, int]],
+        username: str,
+        date: str = None
+) -> None:
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
+        order = Order.objects.create(user=user)
+
+        if date:
+            order.created_at = datetime.strptime(date, "%Y-%m-%d %H:%M")
```
diana-shyrokikh, line 21: I wrote just variable "date" here, and code still works too, so if this changing won't cause errors, You can try do it too)

services/user.py
```diff
@@ -0,0 +1,57 @@
...
+        first_name=first_name,
+        last_name=last_name
+    )
+
+
+def get_user(user_id: int) -> [User, str]:
+    try:
+        return get_user_model().objects.get(id=user_id)
+    except Exception as e:
+        return e
```
diana-shyrokikh, line 26: You can just return the result of get  if we get an error we will see it too without try-except block

services/user.py
```diff
@@ -0,0 +1,57 @@
...
+        username: str = "",
+        password: str = "",
+        email: str = "",
+        first_name: str = "",
+        last_name: str = "",
+) -> [None, str]:
+    try:
+        user = get_user_model().objects.get(id=user_id)
+    except Exception as e:
+        return e
```
diana-shyrokikh, line 40: the same here) 
and You can use here the get_user() function to simplify your code)

diana-shyrokikh (changes requested):

diana-shyrokikh (approved):
Good Job!

db/models.py
```diff
@@ -46,3 +58,63 @@ class MovieSession(models.Model):
...
+            force_insert, force_update, using, update_fields
+        )
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket_row_seat_movie_session"
+            )
+        ]
```
YuriiKindrat, line 116: calss meta should be placed before all ur methods )

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def create_order(
+        tickets: List[Dict[str, int]],
+        username: str,
+        date: str = None
+) -> None:
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
+        order = Order.objects.create(user=user)
```
YuriiKindrat, line 17: no need to create additional variable for user here

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = "",
+        first_name: str = "",
+        last_name: str = "",
```
YuriiKindrat, line 11: try to use kwargs here, also it is better to use None instead of ""

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = "",
+        password: str = "",
+        email: str = "",
+        first_name: str = "",
+        last_name: str = "",
+) -> [None, str]:
+    user = get_user_model().objects.get(id=user_id)
```
YuriiKindrat, line 34: u can use get_user func implemented higher

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(
+        user_id: int,
+        username: str = "",
+        password: str = "",
+        email: str = "",
+        first_name: str = "",
+        last_name: str = "",
```
YuriiKindrat, line 32: try to use kwargs here

YuriiKindrat (changes requested):

Oleg-s-s (approved):
:+1:

PR #391, opened 2023-07-07
implemented tickets-and-orders

mariiahorbova:


db/models.py
```diff
@@ -17,7 +18,7 @@ def __str__(self) -> str:
 
 
 class Movie(models.Model):
-    title = models.CharField(max_length=255)
+    title = models.CharField(max_length=255, db_index=True)
```
AntonShpakovych, line 21: I think its bad practice when you use db_index=True. Because when the model expands it will be difficult to maintain

db/models.py
```diff
@@ -46,3 +47,83 @@ class MovieSession(models.Model):
...
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(to=User, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+    class Meta:
```
AntonShpakovych, line 63: class meta should be after attribute and before methods

db/models.py
```diff
@@ -46,3 +47,83 @@ class MovieSession(models.Model):
...
+
+    def __str__(self) -> str:
+        return (
+            f"{self.movie_session.movie.title} "
+            f"{self.movie_session.show_time} "
+            f"(row: {self.row}, seat: {self.seat})"
+        )
+
+    def clean(self) -> None:
+        from django.core.exceptions import ValidationError
```
AntonShpakovych, line 97: bad practice I think mentors reject if u will use import like this

db/models.py
```diff
@@ -46,3 +47,83 @@ class MovieSession(models.Model):
...
+
+        if not 1 <= self.row <= self.movie_session.cinema_hall.rows:
```
AntonShpakovych, line 99: if not (1 <= self.row <= self.movie_session.cinema_hall.rows):

db/models.py
```diff
@@ -46,3 +47,83 @@ class MovieSession(models.Model):
...
+            raise ValidationError(
+                {
+                    "row": [
+                        f"row number must be in available range: (1, rows): "
+                        f"(1, {self.movie_session.cinema_hall.rows})"
+                    ]
+                }
+            )
+        if not 1 <= self.seat <= self.movie_session.cinema_hall.seats_in_row:
```
AntonShpakovych, line 108: same here

services/movie.py
```diff
@@ -15,13 +17,17 @@ def get_movies(
...
+        queryset = queryset.filter(title__icontains=title)
+
     return queryset
 
 
 def get_movie_by_id(movie_id: int) -> Movie:
     return Movie.objects.get(id=movie_id)
 
 
+@atomic
```
AntonShpakovych, line 30: probably better use transtaction.atomic() but not sure. Mb mentors can help with this)

db/models.py
```diff
@@ -46,3 +47,83 @@ class MovieSession(models.Model):
...
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(to=User, on_delete=models.CASCADE)
```
AntonShpakovych, line 58: related_name

services/movie.py
```diff
@@ -1,11 +1,13 @@
 from django.db.models import QuerySet
+from django.db.transaction import atomic
 
 from db.models import Movie
 
 
 def get_movies(
     genres_ids: list[int] = None,
```
AntonShpakovych, line 8: where u use this genres_ids ?

services/movie_session.py
```diff
@@ -42,3 +44,11 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> List[dict]:
```
AntonShpakovych, line 49: I think bad formatting but not sure

services/user.py
```diff
@@ -0,0 +1,59 @@
+from django.contrib.auth import get_user_model
+
+from settings import AUTH_USER_MODEL
+
+
+def create_user(
```
AntonShpakovych, line 6: u can use here kwargs and check if key in ["first_name".....] - Oleg hint

AntonShpakovych: and setattr()

services/user.py
```diff
@@ -0,0 +1,59 @@
...
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+
+
+def get_user(user_id: int) -> AUTH_USER_MODEL:
+    user = get_user_model()
```
AntonShpakovych, line 30: optional, but zero sense to create variable

services/user.py
```diff
@@ -0,0 +1,59 @@
...
+    return user.objects.get(id=user_id)
+
+
+def update_user(
```
AntonShpakovych, line 34: same like create

db/models.py
```diff
@@ -46,3 +47,80 @@ class MovieSession(models.Model):
...
```
AntonShpakovych, line 58: to=settings.AUTH_USER_MODEL

db/models.py
AntonShpakovych: also related_name

db/models.py
```diff
@@ -17,7 +18,7 @@ def __str__(self) -> str:
 
 
 class Movie(models.Model):
-    title = models.CharField(max_length=255)
+    title = models.CharField(max_length=255, db_index=True)
     description = models.TextField()
     actors = models.ManyToManyField(to=Actor)
```
AntonShpakovych, line 23: related_name

services/user.py
AntonShpakovych: with setattr

AntonShpakovych (changes requested):

services/movie.py
YuriiKindrat: yeah, it is more usual to use @transaction.atomic decorator,but it think this is not critical

YuriiKindrat (approved):

PR #394, opened 2023-07-11
'Solution'

RedCatHot:


db/models.py
```diff
@@ -46,3 +54,66 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE)
```
YuriiKindrat, line 63: bad formmating(

YuriiKindrat (changes requested):
all tests should pass

db/models.py
RedCatHot: image
image

YuriiKindrat (changes requested):
all tests should pass(
delete unused import with Uniqueconstraint

Oleg-s-s (approved):

PR #396, opened 2023-07-14
py-tickets-and-orders

muriakula:


services/movie_session.py
```diff
@@ -25,10 +25,10 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
...
 
 def update_movie_session(
-    session_id: int,
-    show_time: str = None,
-    movie_id: int = None,
-    cinema_hall_id: int = None,
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None,
```
Bohdan-Salamakha, line 31: use typehint like: Optional[YOUR_VALUE_TYPE] = None
for example: list that contains string elements and can be None has typehint Optional[list[str]]

Bohdan-Salamakha (changes requested):

services/movie.py
```diff
@@ -6,8 +9,12 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
```
YukihiroSM, line 11: These should be optional as well

YukihiroSM (changes requested):
Also, consider using black formatter to fix formatting errors, like `'` in migrations instead of `"`
Else looks good to me :)

Bohdan-Salamakha (approved):

PR #399, opened 2023-07-16
Initial implementation

thorium-zero:


db/models.py
```diff
@@ -22,6 +27,11 @@ class Movie(models.Model):
     actors = models.ManyToManyField(to=Actor)
     genres = models.ManyToManyField(to=Genre)
```
LiudmylaKulinchenko, line 28: Specify `related_name` in both fields

LiudmylaKulinchenko: You can find an example in the [checklist](https://github.com/mate-academy/py-taxi-service-startproject/blob/master/checklist.md)

services/movie_session.py
```diff
@@ -42,3 +45,13 @@ def update_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    if not MovieSession.objects.filter(id=movie_session_id).exists():
+        raise ValueError("Movie session does not exist")
+    return list(
+        MovieSession.objects.get(id=movie_session_id)
+        .ticket_set.all()
```
LiudmylaKulinchenko, line 55: No need to use `all` here
And how about specify a related_name in Ticket model?

services/movie_session.py
```diff
@@ -42,3 +45,13 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    if not MovieSession.objects.filter(id=movie_session_id).exists():
+        raise ValueError("Movie session does not exist")
+    return list(
+        MovieSession.objects.get(id=movie_session_id)
```
LiudmylaKulinchenko, line 54: How about to use MovieSession.objects.filter(id=movie_session_id).first() in the first place to avoid requesting db twice?

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+def create_order(
+    tickets: list[dict],
+    username: str,
+    date: datetime = None
+) -> None:
+    user = find_users(username=username).first()
+    if not user:
+        raise ValueError("User does not exist")
+    if len(tickets) < 1:
+        raise ValueError("Unable to create an order without any tickets")
```
LiudmylaKulinchenko, line 19: Remove these checks. You can just try to create an Order in transaction, It will give you an error anyway in such cases. What do you think?

services/user.py
```diff
@@ -0,0 +1,84 @@
...
+    search_creteria = Q()
+    if username:
+        search_creteria &= Q(username=username)
+    if email:
+        search_creteria &= Q(email=email)
+    if first_name:
+        search_creteria &= Q(first_name=first_name)
+    if last_name:
+        search_creteria &= Q(last_name=last_name)
+    return User.objects.filter(search_creteria)
```
LiudmylaKulinchenko, line 53: Do you really need this function? Why don't just use get_user_model and use try/except if you want to handle such error in your way, if you want to handle them?

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🐣☀🌼

PR #400, opened 2023-07-17
Solution

BogdanFSD:


services/movie.py
```diff
@@ -37,4 +42,5 @@ def create_movie(
     if actors_ids:
         movie.actors.set(actors_ids)
 
-    return movie
```
ArsenPidhoretskyi, line 40: Why you remove return?

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+            )
+
+    return order
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
+    order = Order.objects.all()
+
+    if username:
+        order = Order.objects.filter(user__username=username)
```
ArsenPidhoretskyi, line 35:
```suggestion
        order = order.filter(user__username=username)

```

services/user.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def update_user(user_id: int, **kwargs) -> None:
+    user = get_user_model().objects.get(id=user_id)
+
+    for key, value in kwargs.items():
+        if value is not None and value != "":
+            setattr(user, key, value)
+
+    if "password" in kwargs and kwargs["password"]:
```
ArsenPidhoretskyi, line 37: use .get instead of two checks

ArsenPidhoretskyi (changes requested):

services/order.py
```diff
@@ -0,0 +1,37 @@
+from django.contrib.auth import get_user_model
+from db.models import Ticket, Order
+from django.db import transaction
+from django.db.models import QuerySet
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
```
ArsenPidhoretskyi, line 10:
```suggestion
        date: Optional[str] = None

```

ArsenPidhoretskyi (approved):
Good job

PR #401, opened 2023-07-21
solution

OleksandrGrk:


services/user.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def update_user(
+    user_id: int,
+    username: str = None,
+    password: str = None,
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None,
+) -> None:
```
vsmutok, line 38:
```suggestion
def update_user(user_id: int, **kwargs) -> User:
```

also use loop and setattr

vsmutok (changes requested):

vsmutok (approved):
Looks Good 👍

PR #403, opened 2023-07-25
Develop

Oleksiy-Lyashenko:


pavlejviki (changes requested):

db/models.py
```diff
@@ -46,3 +55,78 @@ class MovieSession(models.Model):
...
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        MovieSession,
+        on_delete=models.CASCADE,
+        related_name="movie_session"
```
Dimosphen1, line 78: You should pick up another related_name, e.g. `tickets` so it would be descriptive enough to call from `MovieSession`.

db/models.py
```diff
@@ -46,3 +55,78 @@ class MovieSession(models.Model):
...
+    )
+    order = models.ForeignKey(
+        Order,
+        on_delete=models.CASCADE,
+        related_name="order"
```
Dimosphen1, line 83: You should pick up another related_name, e.g. `tickets` so it would be descriptive enough to call from MovieSession.

services/user.py
```diff
@@ -0,0 +1,61 @@
...
+
+def update_user(
+        user_id: int,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        email: Optional[str] = None,
+        first_name: Optional[str] = None,
+        last_name: Optional[str] = None
+) -> None:
+    user = get_user_model().objects.get(pk=user_id)
```
Dimosphen1, line 44: You could use `get_user` function here.

Dimosphen1 (changes requested):

PR #406, opened 2023-07-31
Solution

vkrasnovyd:


services/order.py
```diff
@@ -0,0 +1,40 @@
...
+from db.models import Order, Ticket
+
+
+def create_order(
+    tickets: list[dict[str]],
+    username: str,
+    date: Optional[str] = None
+) -> Order:
+
+    with transaction.atomic():
```
EdAlekseiev, line 15: Use ` transaction.atomic` as decorator instead of context manager

EdAlekseiev (approved):
Good job. Pay attention to note

services/order.py
vkrasnovyd: Thanks!

PR #407, opened 2023-07-31
Solution

Quiet-Klirik:


services/order.py
```diff
@@ -0,0 +1,33 @@
...
+from db.models import Order, User, Ticket
+
+
+@transaction.atomic
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> Order:
+    user = User.objects.get(username=username)
```
Dimosphen1, line 15: You should use `get_user_model` here instead of importing and using the `User` model directly.

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    movie_session = MovieSession.objects.get(id=movie_session_id)
```
Dimosphen1, line 48: Try to extract the required information here regarding `row` and `seat` from all tickets using `values` or `values_list`. It will execute a query to extract only required values and will not query for objects as a whole.

Dimosphen1 (changes requested):
Several minor changes were requested.

Polyakiv-Andrey (approved):
Good job!

PR #412, opened 2023-08-15
'Solution'

StinGER-ua:


services/movie_session.py
```diff
@@ -26,19 +34,34 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
...
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    session = MovieSession.objects.filter(id=movie_session_id).values(
+        "ticket__row", "ticket__seat"
+    )
+
+    return [
+        {"row": ticket["ticket__row"], "seat": ticket["ticket__seat"]}
+        for ticket in session
+    ]
```
ArsenPidhoretskyi, line 67: Why you need as base MovieSession? You can filter Tickets.objects.filter....

services/user.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    if email:
+        setattr(user, "email", email)
+
+    if first_name:
+        setattr(user, "first_name", first_name)
+
+    if last_name:
+        setattr(user, "last_name", last_name)
```
ArsenPidhoretskyi, line 27: In this case no need to duplicate code use for loop

services/user.py
```diff
@@ -0,0 +1,63 @@
...
+        user.set_password(password)
+
+    if email:
+        user.email = email
+
+    if first_name:
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
```
ArsenPidhoretskyi, line 61: The same code in create method

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #417, opened 2023-08-24
Solution

niksgawrilenko:


services/order.py
```diff
@@ -0,0 +1,39 @@
...
+from db.models import Order, Ticket, MovieSession
+
+
+def create_order(
+    tickets: list[dict],
+    username: datetime,
+    date: datetime | None = None
+) -> None:
+
+    with transaction.atomic():
```
Dimosphen1, line 13: You could use `transaction.atomic` as a function decorator to cover all operations with a function here and in `create_movie`, prefer using it as a context manager in case you need more granular control.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #423, opened 2023-08-30
Tickets and orders

IProskurnytskyi:


services/movie.py
```diff
@@ -6,16 +7,18 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
pavlejviki, line 10: Use Optional type hint, it indicates that a variable can be either of the specified type or None

Correct in all similar cases

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+from db.models import Order, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> None:
+    user = get_user_model().objects.get(username=username)
+    with transaction.atomic():
```
pavlejviki, line 16: Better use @transaction.atomic() decorator

db/models.py
```diff
@@ -46,3 +55,64 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
```
pavlejviki, line 62: Add related_name attribute for all ForeignKey fields

services/movie.py
pavlejviki, line 10:
```suggestion
    title: Optional[str] = None
```

services/movie.py
```diff
@@ -6,16 +7,18 @@
...
 ) -> QuerySet:
-    queryset = Movie.objects.all()
+    if title is not None:
```
pavlejviki, line 12: if title is not provided the filtering should still work by genres and actors

services/movie.py
```diff
@@ -28,13 +31,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
pavlejviki, line 34: Better use @transaction.atomic() decorator

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+        order = Order.objects.create(user=user)
+        if date:
+            order.created_at = date
+            order.save()
+        for ticket in tickets:
+            Ticket.objects.create(
```
pavlejviki, line 22: you can use Django's bulk_create method to create multiple objects in one database query, which can be much faster than calling create() in a loop

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    orders = Order.objects.all()
+    if username:
+        user = get_user_model().objects.get(username=username)
+        orders = orders.filter(user=user)
```
pavlejviki, line 34:
```suggestion
        orders = orders.filter(user__username=username)
```

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    orders = Order.objects.all()
+    if username:
+        user = get_user_model().objects.get(username=username)
```
pavlejviki, line 33: you can filter straight by username

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
```
pavlejviki, line 27: use get_user_model()

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #424, opened 2023-09-04
Develop

VasylTurok:


db/models.py
```diff
@@ -17,7 +22,7 @@ def __str__(self) -> str:
 
 
 class Movie(models.Model):
-    title = models.CharField(max_length=255)
+    title = models.CharField(max_length=255, db_index=True)
```
Dimosphen1, line 25: `title` should not be a db index, there is a possibility that different films will have the same title. Remove this change from a model here.

db/migrations/0004_alter_order_user.py
```diff
@@ -0,0 +1,20 @@
...
+        ('db', '0003_alter_ticket_movie_session_alter_ticket_order'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='order',
+            name='user',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='orders', to=settings.AUTH_USER_MODEL),
+        ),
+    ]
```
Dimosphen1, line 20: Include only one migration with your PR, there is no need to have 3 of them here.

db/models.py
```diff
@@ -46,3 +51,68 @@ class MovieSession(models.Model):
...
+        related_name="orders",
+        on_delete=models.CASCADE
+    )
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        creation_time = self.created_at.strftime("%Y-%m-%d %H:%M:%S")
+        return creation_time
```
Dimosphen1, line 69: There is no need to allocate a variable to store the output of the `strftime` here, return it directly.

services/movie_session.py
```diff
@@ -42,3 +42,18 @@ def update_movie_session(
...
+        id=movie_session_id
+    )
+    result = []
+    for ticket in movie_session.tickets.all():
+        result.append(
+            {
+                "row": ticket.row,
+                "seat": ticket.seat
+            }
+        )
```
Dimosphen1, line 58: There is no need to iterate through all ticket objects here, build your call from the `Ticket` model filtering by movie session id. You could obtain the required information using `values()` via ORM and convert the obtained info to a list in the return statement. This approach would be more efficient.

settings.py
```diff
@@ -24,5 +24,29 @@
...
+    },
+    {
+        "NAME": "django.contrib.auth.password_validation."
+                "CommonPasswordValidator",
+    },
+    {
+        "NAME": "django.contrib.auth.password_validation."
+                "NumericPasswordValidator",
+    },
 ]
```
Dimosphen1, line 52: There is no need to include `AUTH_PASSWORD_VALIDATORS` here, remove it.

Dimosphen1 (changes requested):

pavlejviki (approved):
Well done!

PR #431, opened 2023-09-11
solution

sashabryl:


services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
Dimosphen1, line 36: You could use the atomic transaction as a function decorator here.

Dimosphen1 (changes requested):
Several changes were requested. Besides that, your solution needs to pass all tests before the next check.

Dimosphen1 (approved):
LGTM, well done!

PR #434, opened 2023-09-13
implemented solution

innyshka:


db/models.py
```diff
@@ -22,6 +27,11 @@ class Movie(models.Model):
     actors = models.ManyToManyField(to=Actor)
     genres = models.ManyToManyField(to=Genre)
 
+    class Meta:
+        indexes = [
+            models.Index(fields=["title"]),
+        ]
+
```
OlegatorLE, line 34: if you want to index one column at a time, it might be better to use db_index=True?

db/models.py
```diff
@@ -46,3 +56,79 @@ class MovieSession(models.Model):
...
+        self.full_clean()
+        return super(Ticket, self).save(
+            force_insert,
+            force_update,
+            using,
+            update_fields
+        )
+
+    def __str__(self) -> str:
+        return (f"Matrix {str(self.movie_session.show_time)} "
```
OlegatorLE, line 129: Matrix? what about other movies?

OlegatorLE (changes requested):
check __str__ in Ticket

PR #435, opened 2023-09-13
Solution

OlegatorLE:


RPodolets (approved):

db/models.py
```diff
@@ -46,3 +53,73 @@ class MovieSession(models.Model):
...
+            )
+
+    def save(
+            self,
+            force_insert: bool = False,
+            force_update: bool = False,
+            using: Any = None,
+            update_fields: Any = None) -> None:
+        self.full_clean()
+        return super(Ticket, self).save(
```
Sofbog, line 119: do we need to pass `Ticket, self` into `super()`

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None,
+) -> User:
```
Sofbog, line 12: check [README.md](https://github.com/mate-academy/py-tickets-and-orders#readme) what function supposed to retutn

Sofbog (changes requested):

db/models.py
OlegatorLE: No, I do not need to pass Ticket, self into the super(). TY!

services/user.py
OlegatorLE: fixed annotation

Sofbog (approved):
LGTM

PR #437, opened 2023-09-13
Solution

nyanplague:


db/models.py
```diff
@@ -46,3 +55,86 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    email = models.EmailField()
+    first_name = models.CharField(max_length=255, blank=True)
+    last_name = models.CharField(max_length=255, blank=True)
```
lgSilay, line 63: All this fields already exist in AbstractUser, so i'm not sure that it's was nessesary to redefine it

lgSilay (approved):
Good joob!)

db/models.py
nyanplague: @lgSilay Thank you! I was confused with blank :))

nyanplague:
I believe branch conflict appears because master was changed

PR #438, opened 2023-09-14
'Solution'

lgSilay:


db/models.py
```diff
@@ -22,6 +29,11 @@ class Movie(models.Model):
     actors = models.ManyToManyField(to=Actor)
     genres = models.ManyToManyField(to=Genre)
 
+    class Meta:
+        indexes = [
+            models.Index(fields=["title"])
+        ]
+
```
eduardhabryd, line 36: It's great that you've added indexes here

db/models.py
```diff
@@ -46,3 +58,84 @@ class MovieSession(models.Model):
...
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket_constraint",
+            )
+        ]
```
eduardhabryd, line 97: I think Meta should be placed at the very end of a model

services/movie_session.py
```diff
@@ -42,3 +42,10 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    session = MovieSession.objects.get(id=movie_session_id)
```
eduardhabryd, line 48: You can use get_movie_session_by_id function here in order to not repeat yourself

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
+
+        order = Order.objects.create(user=user)
+        if date is not None:
```
eduardhabryd, line 16: You can just use: `if date:`

services/user.py
```diff
@@ -0,0 +1,54 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = get_user_model().objects.get(id=user_id)
```
eduardhabryd, line 41: you can use get_user function here

eduardhabryd (changes requested):
GJ! but can be improved a little bit

services/movie_session.py
lgSilay: Yeah, it will be better!) Thx:)

lgSilay:
> GJ! but can be improved a little bit
"Thanks for your comment! Your comments really simplified my solution a bit! 😊"

PR #442, opened 2023-09-14
Develop

CaCtUs1k:


services/user.py
```diff
@@ -0,0 +1,45 @@
...
+    return get_user_model().objects.get(pk=user_id)
+
+
+def update_user(
+        user_id: int,
+        username: str = "",
+        password: str = "",
+        email: str = "",
+        first_name: str = "",
+        last_name: str = ""
```
9rosLove, line 32: I think using None would make the code more readable and also would take slightly less memory)

9rosLove (approved):

services/user.py
CaCtUs1k: some tests fail if i using None

CaCtUs1k: fail test_user_service_create_user

CaCtUs1k: i'm sorry, i tried using None on 9-11 lines. My bad

services/user.py
```diff
@@ -0,0 +1,45 @@
...
+
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = "",
+        last_name: str = ""
```
Sofbog, line 11: why you still use empty string instead of `None`

Sofbog (changes requested):

services/user.py
CaCtUs1k: Because  test_user_service_create_user fails if i use None in this lines

PR #443, opened 2023-09-14
Develop

anna-lybid:


db/models.py
```diff
@@ -46,3 +53,71 @@ class MovieSession(models.Model):
...
+    def clean(self) -> None:
+        if not (self.row <= self.movie_session.cinema_hall.rows):
+            raise ValidationError({"row": [
+                f"row number must be in available range: (1, rows): "
+                f"(1, {self.movie_session.cinema_hall.rows})"]})
+        if not (self.seat <= self.movie_session.cinema_hall.seats_in_row):
+            raise ValidationError({"seat": [
+                f"seat number must be in available range: (1, seats_in_row): "
+                f"(1, {self.movie_session.cinema_hall.seats_in_row})"
+            ]})
```
maxkatkalov, line 108: make every element of ]}) on new line, better – use black to reformat yr code

db/models.py
```diff
@@ -46,3 +53,71 @@ class MovieSession(models.Model):
...
+            using: str = None,
+            update_fields: list = None
+    ) -> None:
+        self.full_clean()
+        return super(Ticket, self).save(
+            force_insert=False,
+            force_update=False,
+            using=None,
+            update_fields=None
+        )
```
maxkatkalov, line 123: django docs use the args and kwargs for `save()` and without any argument for `super()`, so we should do the same.

See example [here](https://docs.djangoproject.com/en/4.2/ref/models/instances/#customizing-model-loading).

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
 ) -> QuerySet:
```
maxkatkalov, line 11:
```suggestion
) -> QuerySet:[Movie]
```

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+        for ticket in tickets:
+            Ticket.objects.create(
+                row=ticket["row"],
+                seat=ticket["seat"],
+                movie_session_id=ticket["movie_session"],
+                order=new_order
+            )
+
+
+def get_orders(username: str = None) -> list[Order]:
```
maxkatkalov, line 31:
```suggestion
def get_orders(username: str = None) -> QuerySet[Order]:
```

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
maxkatkalov, line 10: use typing.Optional

services/movie.py
```diff
@@ -28,13 +33,15 @@ def create_movie(
     genres_ids: list = None,
     actors_ids: list = None,
```
maxkatkalov, line 34: use typing.Optional

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+
+from django.db import transaction
+
+from db.models import Ticket, Order
+
+
+def create_order(
+        tickets: list,
+        username: str,
+        date: str = None
```
maxkatkalov, line 11: use typing.Optional

services/user.py
```diff
@@ -0,0 +1,46 @@
...
+    return CustomUser.objects.get(id=user_id)
+
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
```
maxkatkalov, line 32: use typing.Optional

maxkatkalov (changes requested):
Good in general, but some changes are needed

db/models.py
anna-lybid: Thanks!

maxkatkalov (approved):

PR #454, opened 2023-09-15
Solution

OleksandrMoroz24:


db/models.py
```diff
@@ -46,3 +58,85 @@ class MovieSession(models.Model):
...
+            force_update: bool = False,
+            using: Any = None,
+            update_fields: Any = None
+    ) -> None:
+        self.full_clean()
+        return super(Ticket, self).save(
+            force_insert=False,
+            force_update=False,
+            using=None,
+            update_fields=None
```
Sofbog, line 137: Do we need a hardcode here?

services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
```
Sofbog, line 42: You are using  ` QuerySet`  in `def get_movies_sessions(session_date: str = None) -> QuerySet:` but i can’t find import

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                row=ticket["row"],
+                seat=ticket["seat"],
+                movie_session_id=ticket["movie_session"],
+                order=new_order
+            )
+
+        return new_order
```
Sofbog, line 30: check [README.md](https://github.com/mate-academy/py-tickets-and-orders#readme) about function return

db/models.py
```diff
@@ -25,6 +32,11 @@ class Movie(models.Model):
     def __str__(self) -> str:
```
Sofbog, line 32: what about  `related_name` all over the code

Sofbog (changes requested):

services/order.py
OleksandrMoroz24: you mean that this function and create_user in user.py shouldn't return anything?

db/models.py
```diff
@@ -42,11 +54,92 @@ def __str__(self) -> str:
...
+
+    def save(
+            self,
+            force_insert: bool = False,
+            force_update: bool = False,
+            using: Any = None,
+            update_fields: Any = None
+    ) -> None:
+        self.full_clean()
+        return super(Ticket, self).save()
```
Sofbog, line 141: do we need to provide arrangements to` super()` ?, and we definitely need to provide arguments to `save()`

Sofbog (changes requested):

PR #455, opened 2023-09-16
'Solution'

OstapIvakh:


db/models.py
```diff
@@ -19,8 +23,13 @@ def __str__(self) -> str:
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
-    actors = models.ManyToManyField(to=Actor, related_name="movies")
-    genres = models.ManyToManyField(to=Genre, related_name="movies")
+    actors = models.ManyToManyField(to=Actor)
+    genres = models.ManyToManyField(to=Genre)
```
RPodolets, line 27: add related_name

RPodolets (approved):

db/models.py
```diff
@@ -42,11 +51,88 @@ def __str__(self) -> str:
...
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket"
+            )
+        ]
+
+    def clean(self) -> None:
+        cinema_hall = CinemaHall.objects.first()
```
fsocie7y, line 110: i think better to use smth like this hall = self.movie_session.cinema_hall
`couse ticket can be not only for first hall)

services/user.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = User.objects.create_user(
```
fsocie7y, line 12: better to use get_user_model()

services/user.py
```diff
@@ -0,0 +1,58 @@
...
+        user.last_name = last_name
+
+    user.save()
+
+    return user
+
+
+def get_user(user_id: int) -> User | None:
+    try:
+        user = User.objects.get(pk=user_id)
```
fsocie7y, line 31: the same as above

fsocie7y (changes requested):

services/user.py
```diff
@@ -0,0 +1,58 @@
+from db.models import User
+from django.core.exceptions import ObjectDoesNotExist
```
fsocie7y, line 2: fix order imports

services/order.py
```diff
@@ -0,0 +1,39 @@
+from django.contrib.auth import get_user_model
+from django.db import transaction
+from django.utils import timezone
```
fsocie7y, line 3: add blank line between imports with other level

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+
+            Ticket.objects.create(
+                movie_session=movie_session,
+                order=order,
+                row=row,
+                seat=seat,
+            )
+
+
+def get_orders(username: str = None) -> Order:
```
fsocie7y, line 36: fix type annotation, u return a queryset

fsocie7y (changes requested):

PR #459, opened 2023-09-17
solution

Illia-Kononenko:


services/order.py
```diff
@@ -0,0 +1,38 @@
...
+                    order=order,
+                )
+                for ticket in tickets
+            ]
+        )
+
+    return order
+
+
+def get_orders(username: Optional[str] = None) -> Order:
```
ochernous, line 35: Maybe we will get queryset, not single order?

services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
+    return list(
+        Ticket.objects.filter(movie_session_id=movie_session_id).values(
```
ochernous, line 49: You can use previously created function here

db/models.py
```diff
@@ -46,3 +51,71 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
```
ochernous, line 58: Maybe add related_name?

ochernous (approved):

services/user.py
```diff
@@ -0,0 +1,33 @@
...
+from db.models import User
+
+
+def create_user(username: str, password: str, **kwargs) -> None:
+    default_kwargs = {"email": "", "first_name": "", "last_name": ""}
+    kwargs = {**default_kwargs, **kwargs}
+
+    get_user_model().objects.create_user(
+        username=username, password=password, **kwargs
+    )
```
Sofbog, line 14: `README` has a list of parameters that should be allowed in this function

Sofbog (changes requested):

services/user.py
Illia-Kononenko: Sorry, but I don't understand what my mistake is
I created the function according to this tutorial, it has 2 required and 3 optional parameters:
"create_user, takes username, password, optional parameters email, first_name, last_name. The method should save user properly (password should be encrypted), use method .create_user() for this purpose. Also, set fields if appropriate optional values are provided. "

Sofbog (approved):
LGTM

PR #460, opened 2023-09-18
Solution

ilizh:


db/models.py
```diff
@@ -46,3 +54,71 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(null=True)
```
EdAlekseiev, line 60: According to the tack you must set auto_now_add for created_at field to fill this field automatically during the creation

services/movie.py
```diff
@@ -28,13 +35,13 @@ def create_movie(
...
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
+    with transaction.atomic():
```

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def create_order(
+        tickets: list,
+        username: str,
+        date: Optional[str] = None
+) -> Order | None:
+    try:
+        user = get_user_model().objects.get(username=username)
+        with transaction.atomic():
```
EdAlekseiev, line 16: Use `transaction.atomic` as decorator instead of context manager

EdAlekseiev (changes requested):

olena-hul (approved):
Good job!

PR #461, opened 2023-09-18
py-tickets-and-orders solution

KolBohdan:


db/models.py
```diff
@@ -50,3 +55,80 @@ class MovieSession(models.Model):
...
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        to=settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE
+    )
+
```
viktoria-rybenchuk, line 66: add related name

db/models.py
```diff
@@ -50,3 +55,80 @@ class MovieSession(models.Model):
...
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return (str(self.created_at).split("."))[0]
```
viktoria-rybenchuk, line 71:
```suggestion
        return str(self.created_at)
```

viktoria-rybenchuk: This updated code will achieve the same result as your original code, but it's cleaner and more straightforward by directly converting self.created_at to a string without splitting and formatting

db/models.py
```diff
@@ -50,3 +55,80 @@ class MovieSession(models.Model):
...
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        to=MovieSession,
+        on_delete=models.CASCADE
+    )
+    order = models.ForeignKey(to=Order, on_delete=models.CASCADE)
```
viktoria-rybenchuk, line 79: add related name

db/models.py
```diff
@@ -50,3 +55,80 @@ class MovieSession(models.Model):
...
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return (str(self.created_at).split("."))[0]
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
```
viktoria-rybenchuk, line 75: add related name

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Nice!

PR #464, opened 2023-09-19
py-tickets-and-orders

yaroslav-mazurkevych:


services/movie_session.py
```diff
@@ -25,10 +25,10 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
...
 
 def update_movie_session(
-    session_id: int,
-    show_time: str = None,
-    movie_id: int = None,
-    cinema_hall_id: int = None,
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None,
```
Sofbog, line 31: for what reason extra spaces added?

services/user.py
```diff
@@ -0,0 +1,52 @@
...
+
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = "",
+        first_name: str = "",
+        last_name: str = ""
```
Sofbog, line 11: can you follow the same code style like in :
```
def update_user(
        user_id: int,
        username: str = None,
        password: str = None,
        email: str = None,
        first_name: str = None,
        last_name: str = None
) -> None:
```

Sofbog (changes requested):

services/user.py
yaroslav-mazurkevych: Sorry, but if i`ll try to do like this email: str = None,  first_name: str = None, last_name: str = None, my code won`t work)

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+) -> None:
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
+        order = Order.objects.create(user=user)
+
+        if date:
+            order.created_at = date
+            order.save()
+
+        for ticket_data in tickets:
```
Kroll410, line 20: You should not use create method in a loop - consider using bulk create method

Kroll410 (changes requested):

services/order.py
yaroslav-mazurkevych: bulk_create handle exception and don`t create ticket

PR #468, opened 2023-09-27
Solution tickets and orders

MarkShapka:


db/models.py
```diff
@@ -50,3 +59,64 @@ class MovieSession(models.Model):
...
+                {"row": [f"row number must be in available range: "
+                         f"(1, rows): (1, "
+                         f"{self.movie_session.cinema_hall.rows})"]}
+            )
+        if not 1 < self.seat <= self.movie_session.cinema_hall.seats_in_row:
+            raise ValidationError(
+                {"seat": [f"seat number must be in available range: "
+                          f"(1, seats_in_row): (1, "
+                          f"{self.movie_session.cinema_hall.seats_in_row})"]}
+            )
```
Dimosphen1, line 107: You could avoid code duplication here because there are several repetitive parts in this method.

Dimosphen1 (approved):

Dimosphen1 (changes requested):

db/migrations/0001_initial.py
```diff
@@ -42,17 +73,46 @@ class Migration(migrations.Migration):
...
+        ),
+        migrations.AddConstraint(
+            model_name='ticket',
+            constraint=models.UniqueConstraint(fields=('row', 'seat', 'movie_session'), name='uniq_ticket'),
+        ),
+        migrations.AddIndex(
+            model_name='movie',
+            index=models.Index(fields=['title'], name='movie_title_idx'),
         ),
     ]
```
Dimosphen1, line 118: Do not modify the migration that is already present in a file. Create a new one instead and leave the first one unmodified.

services/movie.py
```diff
@@ -28,13 +33,15 @@ def create_movie(
...
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    movie = None
+    with transaction.atomic():
```
Dimosphen1, line 37: You could use the `transaction.atomic` decorator for this method.

services/movie_session.py
```diff
@@ -42,3 +43,10 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    seats = list(Ticket.objects.filter(movie_session_id=movie_session_id)
+                 .values("row", "seat"))
```
Dimosphen1, line 50: Do not break line before `.`, it would be better to make line breaking after the parenthesis.

services/user.py
```diff
@@ -0,0 +1,57 @@
...
+def get_user(user_id: int) -> User:
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> User:
```
Dimosphen1, line 37: Indicate that these parameters are optional in type annotation, e.g.
```suggestion
                last_name: str | None = None) -> User:
```
Apply this change where applicable.

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
GJ!

PR #475, opened 2023-10-07
Solution2

mwellick:


services/user.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+    if email:
+        user.email = email
+
+    if first_name:
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
```
ArsenPidhoretskyi, line 23: The same code in update_user
Move it to separate function

ArsenPidhoretskyi (changes requested):

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+from django.contrib.auth import get_user_model
+from django.db.models import QuerySet
+from django.db import transaction
+from db.models import Order, Ticket
+
+
+def create_order(tickets: list[dict],
+                 username: str,
+                 date: Optional[str] = None) -> QuerySet:
+    with transaction.atomic():
```
EdAlekseiev, line 11: Use `transaction.atomic` as decorator instead of context manager. It makes you code cleaner

EdAlekseiev (changes requested):

YukihiroSM (approved):
Looks good

PR #476, opened 2023-10-08
Solution

Artemh59:


db/migrations/0001_initial.py
```diff
@@ -42,17 +73,46 @@ class Migration(migrations.Migration):
...
+            ],
+        ),
+        migrations.AddConstraint(
+            model_name='ticket',
+            constraint=models.UniqueConstraint(fields=('row', 'seat', 'movie_session'), name='unique_tickets'),
+        ),
+        migrations.AddIndex(
+            model_name='movie',
+            index=models.Index(fields=['title'], name='db_movie_title_5d0841_idx'),
         ),
```
Dimosphen1, line 117: Create a new migration, do not rewrite an existing one.

services/movie.py
```diff
@@ -6,6 +9,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
```
Dimosphen1, line 11: All arguments should be optional for this function.

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+        tickets: list[dict],
+        username: str,
+        date: Optional[str] = None
+) -> None:
+    user = User.objects.get(username=username)
+    order = Order(user=user)
+
+    if date:
+        order.created_at = date
+    order.save()
```
Dimosphen1, line 20: Perform order save only when the `date` argument is present.

Dimosphen1 (changes requested):

db/migrations/0001_initial.py
Dimosphen1: The comment was not implemented.

Dimosphen1 (changes requested):
Create a new migration and do not overwrite the existing one.

pavlejviki (approved):
Good job!

PR #479, opened 2023-10-12
Solution

AlexPelevin99:


db/migrations/0003_user_order_ticket_movie_db_movie_title_5d0841_idx_and_more.py
```diff
@@ -0,0 +1,94 @@
...
+        migrations.AddField(
+            model_name='user',
+            name='user_permissions',
+            field=models.ManyToManyField(blank=True, help_text='Specific permissions for this user.', related_name='user_set', related_query_name='user', to='auth.Permission', verbose_name='user permissions'),
+        ),
+        migrations.AddConstraint(
+            model_name='ticket',
+            constraint=models.UniqueConstraint(fields=('row', 'seat', 'movie_session'), name='unique_row_seat_movie_session'),
+        ),
+    ]
```
Dimosphen1, line 94: It is a good practice to include one migration per app per pull request in Django projects. Push only one migration.

services/user.py
```diff
@@ -0,0 +1,59 @@
...
+
+from db.models import User
+
+
+def create_user(
+    username: str,
+    password: str,
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None,
```
Dimosphen1, line 11: When you have default values, indicate that your arguments are optional in type annotation. Apply this change throughout your code
```suggestion
    email: str | None = None,
    first_name: str | None = None,
    last_name: str | None = None,
```

services/user.py
```diff
@@ -0,0 +1,59 @@
...
+) -> User:
+    with transaction.atomic():
```
Dimosphen1, line 13: You could apply a `transaction.atomic` decorator to this function so that all function body would be treated as atomic transaction.

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Ticket, Order, User
+
+
+def create_order(tickets: list, username: str, date: str = None) -> Order:
+    ticket_objects: list[Ticket] = []
+    with transaction.atomic():
```
viktoria-rybenchuk, line 11: use `@transaction.atomic`

services/user.py
```diff
@@ -0,0 +1,59 @@
+from django.db import transaction
+
+from db.models import User
+
+
+@transaction.atomic
```
viktoria-rybenchuk, line 6: why do you need transaction here?

viktoria-rybenchuk (changes requested):

services/user.py
AlexPelevin99: mentor give me this advise to use transaction decorator on this method

Dimosphen1: @AlexPelevin99 my previous comment did not take into account the underlying logic of the function, thanks @viktoria-rybenchuk for pointing this out. Comment from @viktoria-rybenchuk makes sense because using the `transaction.atomic` makes more sense when there are transactions that involve several objects and the whole transaction makes sense only when all changes are committed to the DB, otherwise, it will fall back to an initial state. Here you perform an operation over only one newly created object, which is why explicitly declaring that this transaction should be atomic is too cautious. I will approve this PR, however, do not overuse this decorator and prefer it when transactions include multiple objects and they depend on each other.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

services/user.py
AlexPelevin99: @Dimosphen1 @viktoria-rybenchuk ok, thank you, for your comments

PR #480, opened 2023-10-12
Develop

3lobites:


services/movie.py
```diff
@@ -28,13 +35,14 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+
+from db.models import Order, MovieSession, User, Ticket
+
+
+def create_order(
+        tickets: List[dict],
+        username: str,
+        date: Optional[str] = None
+) -> Order:
+    with transaction.atomic():
```

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+        if User.objects.filter(username=username):
+
+            new_order = Order.objects.create(
+                user=User.objects.get(username=username)
+            )
```
EdAlekseiev, line 19: Don't make two of the same request to db. You can store the result of `User.objects.filter(username=username)` and use it where you need.

EdAlekseiev (changes requested):

services/movie.py
```diff
@@ -1,11 +1,15 @@
...
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+    genres_ids: List[int] = None,
+    actors_ids: List[int] = None,
+    title: Optional[str] = None
```
EdAlekseiev, line 12:
```suggestion
    genres_ids: Optional[List[int]] = None,
    actors_ids: Optional[List[int]] = None,
    title: Optional[str] = None
```

EdAlekseiev (changes requested):

3lobites:
Optional prescribed according to recommendations

services/movie_session.py
```diff
@@ -42,3 +44,10 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> List[dict]:
+
+    work_cinema_hall = MovieSession.objects.get(id=movie_session_id)
```
ArsenPidhoretskyi, line 51: Why you start query from MovieSession, instead use Ticket

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+from db.models import Order, MovieSession, User, Ticket
+
+
+@transaction.atomic
+def create_order(
+        tickets: List[dict],
+        username: str,
+        date: Optional[str] = None
+) -> Order:
+    work_user = User.objects.filter(username=username)
```
ArsenPidhoretskyi, line 15: Use .first()

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+
+from db.models import Order, MovieSession, User, Ticket
+
+
+@transaction.atomic
+def create_order(
+        tickets: List[dict],
+        username: str,
+        date: Optional[str] = None
+) -> Order:
```
ArsenPidhoretskyi, line 14: It can return none, if user was not found

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+            new_order.created_at = date_time
+            new_order.save()
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                order=new_order,
+                seat=ticket["seat"],
+                row=ticket["row"],
+                movie_session=MovieSession.objects.get(
+                    id=ticket["movie_session"]
```
ArsenPidhoretskyi, line 31: Use movie_session_id=

services/user.py
```diff
@@ -0,0 +1,58 @@
...
+
+    user_upd = User.objects.filter(id=user_id)
+    if username:
+        user_upd.update(username=username)
+    if email:
+        user_upd.update(email=email)
+    if first_name:
+        user_upd.update(first_name=first_name)
+    if last_name:
+        user_upd.update(last_name=last_name)
```
ArsenPidhoretskyi, line 53: So much query.... Set all variables and then just save

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):
Good job

PR #482, opened 2023-10-15
Solution

dmitriy-kds:


services/movie_session.py
```diff
@@ -42,3 +44,8 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> List[dict]:
+    return MovieSession.objects.filter(id=movie_session_id).\
+        values("ticket__row", "ticket__seat")
```
EdAlekseiev, line 51: You must return list of dict instead of QuerySet

EdAlekseiev (changes requested):

PR #484, opened 2023-10-19
implemented py-tickets-and-orders task

HlovaOleh:


services/movie_session.py
```diff
@@ -24,6 +24,17 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
     return MovieSession.objects.get(id=movie_session_id)
 
 
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    session = MovieSession.objects.get(id=movie_session_id)
+    tickets = Ticket.objects.filter(movie_session=session)
```
ArsenPidhoretskyi, line 29:
```suggestion
    tickets = Ticket.objects.filter(movie_session_id=movie_session_id)
```
No need to do two queries

services/movie_session.py
```diff
@@ -24,6 +24,17 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
...
+
+    taken_seats = [
+        {"row": ticket.row, "seat": ticket.seat} for ticket in tickets
+    ]
+
+    return taken_seats
```
ArsenPidhoretskyi, line 35: Also you can use .values for ORM

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+
+        return new_order
+
+
+def get_orders(username: Optional[str] = None) -> QuerySet[Order]:
+    if username:
+        user = get_user_model().objects.get(username=username)
+        return Order.objects.filter(user=user)
+
+    return Order.objects.all()
```
ArsenPidhoretskyi, line 42:
```suggestion
    orders = Order.objects.all()
    if username:
        user = get_user_model().objects.get(username=username)
        orders = orders.filter(user=user)

    return orders
```
This will help support your code in future

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+                seat=ticket["seat"]
+            )
+
+        return new_order
+
+
+def get_orders(username: Optional[str] = None) -> QuerySet[Order]:
+    if username:
+        user = get_user_model().objects.get(username=username)
+        return Order.objects.filter(user=user)
```
ArsenPidhoretskyi, line 40: Also can you try like this:
```suggestion
        return Order.objects.filter(user__username=username)
```

services/user.py
```diff
@@ -0,0 +1,63 @@
...
+        user.set_password(password)
+
+    if email:
+        user.email = email
+
+    if first_name:
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
```
ArsenPidhoretskyi, line 61: The same code into create_user

ArsenPidhoretskyi (changes requested):

settings.py
```diff
@@ -6,6 +6,15 @@
 # SECURITY WARNING: Modify this secret key if using in production!
 SECRET_KEY = "6few3nci_q_o@l1dlbk81%wcxe!*6r29yu629&d97!hiqat9fa"
 
+INSTALLED_APPS = [
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    # "django_extensions",
```
YukihiroSM, line 12: Remove commented code if it's unused

YukihiroSM (changes requested):

PR #490, opened 2023-10-28
Solution

endernaga:


db/migrations/0001_initial.py
EdAlekseiev, line 1: Why did you delete migration files?

db/models.py
```diff
@@ -50,3 +58,58 @@ class MovieSession(models.Model):
...
+        if not (1 <= self.seat <= self.movie_session.cinema_hall.seats_in_row):
+            raise ValidationError({
+                "seat": f"seat number must be in available range:"
+                f" (1, seats_in_row):"
+                f" (1, {self.movie_session.cinema_hall.seats_in_row})"})
+
+    def save(self, force_insert: bool = False,
+             force_update: bool = False,
+             using: None = None,
+             update_fields: None = None) -> QuerySet:
```
EdAlekseiev, line 111: Each argument must be placed on a separate line

EdAlekseiev, line 111:
```suggestion
    def save(
        self, 
        force_insert: bool = False,
        force_update: bool = False,
        using: None = None,
        update_fields: None = None
    ) -> QuerySet:
```

services/movie.py
```diff
@@ -28,13 +34,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
EdAlekseiev, line 37: Use `transaction.atomic` as decorator instead of context manager

services/order.py
```diff
@@ -0,0 +1,30 @@
...
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Ticket, Order
+
+
+def create_order(tickets: List[dict],
+                 username: str,
+                 date: datetime.datetime = None) -> None:
+    with transaction.atomic():
```

EdAlekseiev (changes requested):

db/migrations/0001_initial.py
endernaga: my db baged , and i found that solution in web

ArsenPidhoretskyi, line 1: Restore this file

ArsenPidhoretskyi (changes requested):

NikOneZ1 (approved):
Good job!

PR #491, opened 2023-10-29
f

XOctOpus1:


vsmutok (changes requested):
what does this `f` mean?
please create normal commit names

image

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+from db.models import User, Ticket, Order, MovieSession
+
+
+def create_order(
+        tickets: list[Ticket],
+        username: str,
+        date: str = None
+) -> Order:
+
+    with transaction.atomic():
```
vsmutok, line 12: You could use the transaction.atomic decorator for this method.

services/movie.py
```diff
@@ -28,13 +32,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
vsmutok, line 35: You could use the transaction.atomic decorator for this method.

vsmutok (changes requested):

db/models.py
```diff
@@ -50,3 +59,78 @@ class MovieSession(models.Model):
...
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        null=True
+    )
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
ArsenPidhoretskyi, line 84:
```suggestion
        return str(self.created_at)
```
In this case better to use str

services/user.py
```diff
@@ -0,0 +1,59 @@
...
+    return user
+
+
+def get_user(user_id: int) -> User:
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(
+        user_id: int,
+        username: str = None,
```
ArsenPidhoretskyi, line 36:
```suggestion
        username: Optional[str] = None,
```
Because None as default. Fix in all places

ArsenPidhoretskyi (changes requested):

db/models.py
XOctOpus1: image

XOctOpus1: коли змінив з'явилась така помилка

XOctOpus1: а ні, то моя вина, забув забрати фігурні лапки

db/migrations/0002_user_order_ticket_alter_movie_actors_and_more.py
```diff
@@ -0,0 +1,113 @@
...
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('last_login', models.DateTimeField(blank=True, null=True, verbose_name='last login')),
+                ('is_superuser', models.BooleanField(default=False, help_text='Designates that this user has all permissions without explicitly assigning them.', verbose_name='superuser status')),
+                ('is_staff', models.BooleanField(default=False, help_text='Designates whether the user can log into this admin site.', verbose_name='staff status')),
+                ('is_active', models.BooleanField(default=True, help_text='Designates whether this user should be treated as active. Unselect this instead of deleting accounts.', verbose_name='active')),
+                ('date_joined', models.DateTimeField(default=django.utils.timezone.now, verbose_name='date joined')),
+                ('username', models.CharField(max_length=68, unique=True)),
+                ('password', models.CharField(max_length=68)),
+                ('email', models.EmailField(max_length=254)),
```
Yurnerroo, line 29: change ' to " with Black

Yurnerroo (changes requested):

Yurnerroo (approved):

PR #493, opened 2023-10-30
Develop

grungezz:


services/movie.py
```diff
@@ -1,11 +1,13 @@
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+        genres_ids: list[int] = None,
```
Sofbog, line 8: fix indentation  all over the code

services/movie_session.py
```diff
@@ -42,3 +42,9 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
```
Sofbog, line 44: you already have function to get MovieSession by id DRY

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+
+from db.models import Order, Ticket
+
+
+@transaction.atomic()
+def create_order(
+        tickets: list,
+        username: str,
+        date: datetime = None
+) -> None:
```
Sofbog, line 15: we need to make sure both models will be created  `transaction.atomic`

Sofbog (changes requested):

Sofbog (approved):
LGTM

PR #494, opened 2023-10-30
Develop

MarinaKhrunina:


services/user.py
```diff
@@ -0,0 +1,64 @@
+from django.contrib.auth import get_user_model
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
```
Grayder0152, line 8: add Optional typing

services/user.py
```diff
@@ -0,0 +1,64 @@
...
+        first_name: str = "",
+        last_name: str = "",
+) -> User:
+
+    user = get_user_model().objects.create_user(
+        username=username,
+        password=password
+    )
+
+    if email:
```
Grayder0152, line 18:
```suggestion
    if email is not None:

```

services/user.py
```diff
@@ -0,0 +1,64 @@
...
+    return get_user_model().objects.get(pk=user_id)
+
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None,
```
Grayder0152, line 42: add Optional typing

Grayder0152 (changes requested):

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+from typing import List
+
+
+def create_order(
+        tickets: List[dict],
+        username: str,
+        date: datetime = None
+) -> None:
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
```
vsmutok, line 17: I think  better to use transaction.atomic decorator

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
vsmutok, line 36: I think better to use transaction.atomic decorator

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Good job!

PR #495, opened 2023-10-30
Solution

Starchild666:


services/order.py
```diff
@@ -0,0 +1,32 @@
...
+                movie_session_id=ticket_data.get("movie_session"),
+                order=order,
+            )
+        return order
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
+    if username:
+        user = get_user_model().objects.get(username=username)
+        return user.order_set.all()
```
ArsenPidhoretskyi, line 31: In this case, you do 2 queries. Optimize them use only one, filter by fk field

ArsenPidhoretskyi (changes requested):

Yurnerroo (approved):

PR #496, opened 2023-10-30
Test(py-tickets-and-orders)

baranovr:


baranovr:
Can you give me a few tips what can I do with it:
error

viktoria-rybenchuk (changes requested):
Try to uninstall Django and then install it again. If that doesn't resolve the issue, please come to the Q&A session with this question, as we will need to check your settings and examine the full traceback

PR #499, opened 2023-11-03
task completed (1)

DaniilHinzburh:


services/order.py
```diff
@@ -0,0 +1,39 @@
...
+            order=order,
+            row=row,
+            seat=seat,
+        )
+
+
+def get_orders(username: str = None) -> List[Order]:
+    if username:
+        return Order.objects.filter(user__username=username)
+    else:
```
vsmutok, line 38: remove redundant  `else`

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+        user.last_name = last_name
+    user.save()
+    return user
+
+
+def get_user(user_id: int) -> User | None:
+    try:
+        return User.objects.get(id=user_id)
+    except User.DoesNotExist:
+        return None
```
vsmutok, line 26: Use the `User.objects.filter` method instead of `User.objects.get` because you're looking for a user by ID, and a user with that ID might not exist. This will avoid you throwing exceptions.

```suggestion
     return User.objects.filter(id=user_id).first()
```

vsmutok (changes requested):

services/order.py
vasylhnatiuk: do it

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+
+        Ticket.objects.create(
+            movie_session=movie_session,
+            order=order,
+            row=row,
+            seat=seat,
+        )
+
+
+def get_orders(username: str = None) -> List[Order]:
```
vasylhnatiuk, line 35:
```suggestion
def get_orders(username: str = None) ->Queryset[Order]:
```

services/user.py
```diff
@@ -0,0 +1,48 @@
...
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
+
+
+def get_user(user_id: int) -> User | None:
+    return User.objects.filter(id=user_id).first()
```
vasylhnatiuk, line 23:
```suggestion
    return User.objects.get(id=user_id)
```

vasylhnatiuk (changes requested):

services/order.py
vsmutok: remove redundant  else

vsmutok (changes requested):

db/models.py
```diff
@@ -39,6 +41,17 @@ def __str__(self) -> str:
         return self.name
 
 
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey("User", on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
ArsenPidhoretskyi, line 49: Use str() method

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+
+from django.db import transaction
+from db.models import Order, Ticket, MovieSession
+from db.models import User
+from django.db.models.query import QuerySet
+
+
+@transaction.atomic
+def create_order(
+        tickets: List[Ticket],
```
ArsenPidhoretskyi, line 11: Do not true it accepts list of dict

services/user.py
```diff
@@ -0,0 +1,48 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None) -> User | None:
+    try:
+        user = User.objects.get(id=user_id)
```
ArsenPidhoretskyi, line 34: Instead of try-except use get_object_or_404

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #504, opened 2023-11-07
Solution

mykolamateichuk:


services/movie_session.py
```diff
@@ -42,3 +42,19 @@ def update_movie_session(
...
+
+    for ticket in tickets:
+        taken_seats.append(
+            {
+                "row": ticket.row,
+                "seat": ticket.seat
+            }
+        )
+
+    return taken_seats
```
Lyutillis, line 60: You cna use list comprehension here

Lyutillis (approved):

db/models.py
```diff
@@ -50,3 +53,61 @@ class MovieSession(models.Model):
...
+                                  f"(1, seats_in_row): (1, {cinema_seats})")
+            raise ValidationError(
+                {"seat": seat_error_message}
+            )
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+    class Meta:
```
LiudmylaKulinchenko, line 107: class Meta should be placed before methods

db/models.py
```diff
@@ -50,3 +53,61 @@ class MovieSession(models.Model):
...
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.created_at.strftime('%Y-%m-%d %H:%M:%S')}"
+
+    class Meta:
```
LiudmylaKulinchenko, line 69: class Meta should be placed before methods

services/movie_session.py
LiudmylaKulinchenko: You can try it, also a good solution. But it's ok to leave regular loop if you think it's more clear

db/models.py
LiudmylaKulinchenko: read checklist, please

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+from db.models import User
+
+
+def create_user(username: str,
+                password: str,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> User:
+
+    user = get_user_model().objects.create(username=username)
```
LiudmylaKulinchenko, line 12: `The method should save user properly (password should be encrypted), use method .create_user() for this purpose`
Read checklist again, please 🙌

LiudmylaKulinchenko (changes requested):

PR #506, opened 2023-11-07
Develop

Familenko:


db/models.py
```diff
@@ -50,3 +55,76 @@ class MovieSession(models.Model):
...
+
+    def __str__(self) -> str:
+        return (f"{self.movie_session.movie.title} "
+                f"{self.movie_session.show_time} "
+                f"(row: {self.row}, seat: {self.seat})")
+
+    def clean(self) -> None:
+
+        if self.row not in range(
+                1, self.movie_session.cinema_hall.rows + 1
```
prochigor, line 95: You can make a variable for "self.movie_session.cinema_hall.rows", you use that 2 times and the condition will be shortest, in one line, but it is up to you

db/models.py
```diff
@@ -50,3 +55,76 @@ class MovieSession(models.Model):
...
+                {"row": [
+                    "row number must be in available range:"
+                    " (1, rows): (1, {})".format
+                    (self.movie_session.cinema_hall.rows)
+                ]
+                }
+            )
+
+        if self.seat not in range(
+                1, self.movie_session.cinema_hall.seats_in_row + 1
```
prochigor, line 108: the same

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+
+    return user
```
prochigor, line 51: There isn't written to return something.

prochigor (changes requested):

db/models.py
```diff
@@ -50,3 +55,77 @@ class MovieSession(models.Model):
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User,
+                             related_name="orders",
+                             on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+    class Meta:
+        ordering = ["-created_at"]
```
LiudmylaKulinchenko, line 74: class Meta should be placed before methods. Don't forget to read checklist, please

LiudmylaKulinchenko: fix everywhere

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+            movie_session = MovieSession.objects.get(
+                id=ticket["movie_session"]
+            )
+            Ticket.objects.create(row=ticket["row"],
+                                  seat=ticket["seat"],
+                                  movie_session=movie_session,
+                                  order=order)
+        if date:
+            order.created_at = date
+            order.save()
```
LiudmylaKulinchenko, line 26: created_at generates automatically. No need in this condition
Also now you do save() only if there is a date. You should do this anyway

LiudmylaKulinchenko (changes requested):

services/order.py
Familenko: ```This method should create an order with the provided user, if date is provided, set created_at to this date```
So my class create datetime automatically
```class Order(models.Model):
    created_at = models.DateTimeField(auto_now_add=True)```
but I rewrite it with this statement
```        if date:
            order.created_at = date
            order.save()
```
in case if I have date as argument. If date=None instance form here ```order = Order.objects.create(user=user)``` will not required save() and will be ignored and order will be just created by method Order.objects.create(user=user) witch is not required save()

db/models.py
```diff
@@ -50,3 +55,77 @@ class MovieSession(models.Model):
...
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User,
```
fsocie7y, line 66: Use get_user_model() instead of User. Check out the [checklist](https://github.com/mate-academy/py-movie-moviesession-cinemahall/blob/master/checklist.md#code-efficiency).

fsocie7y: Or import AUTH_USER_MODEL from settings.py

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+from db.models import User, Order, MovieSession, Ticket
+
+
+def create_order(tickets: list[dict],
+                 username: str,
+                 date: str = None) -> None:
+
+    with transaction.atomic():
+
+        user = User.objects.get(username=username)
```
fsocie7y, line 13: Same about User. Best way is to use get_user_model() instead of User))

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+                                  order=order)
+        if date:
+            order.created_at = date
+            order.save()
+
+
+def get_orders(username: str = None) -> QuerySet:
+
+    if username:
+        user = User.objects.get(username=username)
```
fsocie7y, line 32: Same about User)

services/user.py
```diff
@@ -0,0 +1,49 @@
+from db.models import User
+
+
+def create_user(username: str,
```
fsocie7y, line 4: fix formatting

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+    return User.objects.get(id=user_id)
+
+
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None
+                ) -> None:
```
fsocie7y, line 34: formatting here also should be fixed

services/user.py
fsocie7y: and rewrite usages of User everywhere in according to checklis.md (get_user_model())

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #507, opened 2023-11-08
Solution

MaksymUK:


EdAlekseiev (changes requested):
Fix tests

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
EdAlekseiev, line 36: Use `transaction.atomic` as decorator instead of decorator. It makes your code more readable

services/movie.py
```diff
@@ -1,14 +1,19 @@
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
+    title: str = None,
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
```
EdAlekseiev, line 10: If argument can be None you have to show it in type hint. Fix it here and in other functions
```suggestion
    title: str | None = None,
    genres_ids: list[int] | None = None,
    actors_ids: list[int] | None = None,
```

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+from django.db import transaction
+from db.models import Order, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: User,
+        date: str = None
+) -> None:
+    with transaction.atomic():
```
EdAlekseiev, line 13: Use transaction.atomic as decorator instead of decorator. It makes your code more readable

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+                movie_session_id=ticket["movie_session"],
+                order=order,
+            )
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
+    if username:
+        return Order.objects.filter(user__username=username)
+    else:
+        return Order.objects.all()
```
EdAlekseiev, line 34: You can simplify this code
```suggestion
    orders = Order.objects.all()
    if username:
        orders = orders.filter(user__username=username)
    return orders
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #508, opened 2023-11-08
Solution

prochigor:


services/user.py
```diff
@@ -0,0 +1,51 @@
...
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = get_user_model().objects.get(id=user_id)
+    if username is not None:
```
Shatrovskyi, line 40: U can also use "if username:" rather than "if username is not None", like in create_user() function. In my opinion we have to follow one style rule in our code.

Shatrovskyi (approved):

db/models.py
```diff
@@ -50,3 +58,70 @@ class MovieSession(models.Model):
...
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey("User", on_delete=models.PROTECT)
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+    class Meta:
+        ordering = ["-created_at"]
```
LiudmylaKulinchenko, line 71: class Meta should be placed before methods, don't forget to read checklist, please

LiudmylaKulinchenko: fix everywhere

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return [
+        {"row": item["ticket__row"], "seat": item["ticket__seat"]}
+        for item in MovieSession.objects.filter(
+            ticket__movie_session=movie_session_id
+        ).values("ticket__row", "ticket__seat")
```
LiudmylaKulinchenko, line 52: It's fine, as for me, but you also can filter MovieSessions outside of the return, create another variable to increase readability

LiudmylaKulinchenko (changes requested):

db/models.py
prochigor: In the checklist won't written about Meta should be placed before methods, there written "Implement class Meta only after you defined all attributes", maybe it confused me

PR #510, opened 2023-11-08
Solution

OlehShumov:


Naz-iv (approved):

db/models.py
```diff
@@ -50,3 +58,70 @@ class MovieSession(models.Model):
...
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey("User", on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+    class Meta:
```
LiudmylaKulinchenko, line 70: class Meta should be placed before methods, don't forget to read checklist, please

LiudmylaKulinchenko: fix everywhere

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return [
+        {"row": item["ticket__row"], "seat": item["ticket__seat"]}
+        for item in MovieSession.objects.filter(
+            ticket__movie_session=movie_session_id
+        ).values("ticket__row", "ticket__seat")
```
LiudmylaKulinchenko, line 52: it's fine, as for me, but you also can create additional variable to increase readability

services/user.py
```diff
@@ -0,0 +1,47 @@
...
+    if username is not None:
+        user.username = username
+    if password:
+        user.set_password(password)
+    if email is not None:
+        user.email = email
+    if first_name is not None:
+        user.first_name = first_name
+    if last_name is not None:
+        user.last_name = last_name
```
LiudmylaKulinchenko, line 45: Don't add `is not None` here. Make consistent code style

LiudmylaKulinchenko (changes requested):

PR #513, opened 2023-11-08
Solution

petrykivd:


fedorov-s-od (approved):
Nice job, mate! )

db/models.py
```diff
@@ -50,3 +57,54 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(on_delete=models.CASCADE, to="User")
```
LiudmylaKulinchenko, line 64: Import user model from settings.py

db/models.py
```diff
@@ -50,3 +57,54 @@ class MovieSession(models.Model):
...
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+    class Meta:
+        ordering = ["-created_at"]
```
LiudmylaKulinchenko, line 70: class Meta should be placed before methods. Don't forget to read checklist, please

LiudmylaKulinchenko: Fix everywhere

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
+    return [
+        {"row": item["ticket__row"], "seat": item["ticket__seat"]}
+        for item in MovieSession.objects.filter(
+            ticket__movie_session=movie_session_id
+        ).values("ticket__row", "ticket__seat")
```
LiudmylaKulinchenko, line 52: it's fine, as for me, but you could also create additional variable to increase readability

services/order.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@transaction.atomic
+def create_order(tickets: list,
+                 username: str,
+                 date: str = None) -> None:
+    user = get_user_model().objects.get(username=username)
+    order = Order.objects.create(user_id=user.id)
+    if date:
+        order.created_at = date
```
LiudmylaKulinchenko, line 14: Date creates automatically, you have auto_now_add in the model

services/user.py
```diff
@@ -0,0 +1,48 @@
...
+
+from db.models import User
+
+
+def create_user(username: str,
+                password: str,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None,
+                ) -> None:
```
LiudmylaKulinchenko, line 11: fix formatting

services/order.py
petrykivd: It's rule in task, also if  I delete it from here, 1 test is will be failed

fsocie7y (approved):
GJ 👍

PR #514, opened 2023-11-08
'Solution'

fedorov-s-od:


DenPrislipskyi (approved):

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+
+def create_user(
+        username: str,
+        password: str,
+        email: Optional[str] = None,
+        first_name: Optional[str] = None,
+        last_name: Optional[str] = None
+) -> None:
+    hashed_password = make_password(password)
+    user = User.objects.create_user(
```
LiudmylaKulinchenko, line 16: Use get_user_model() instead of User. Check out the [checklist](https://github.com/mate-academy/py-movie-moviesession-cinemahall/blob/master/checklist.md#code-efficiency).

LiudmylaKulinchenko (changes requested):

services/user.py
fedorov-s-od: I dunno which of these two options is more appropriate:
User = get_user_model()
user = User.objects...
or
user = get_user_model().objects...
:)

fsocie7y (changes requested):
u should use get_user_model() in each case when u want to call User model)

PR #517, opened 2023-11-08
Solution

Daniil-Pankieiev:


SevKrok (approved):
OK

db/models.py
```diff
@@ -50,3 +59,75 @@ class MovieSession(models.Model):
...
+        to=AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+    class Meta:
+        ordering = ["-created_at"]
```
fsocie7y, line 76: class Meta should be placed before methods. fix it everywhere

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+
+
+def get_user(user_id: int) -> User:
```
fsocie7y, line 27: u can use `get_user_model` for type annotation too)

fsocie7y (changes requested):

services/user.py
Daniil-Pankieiev: Fixed

fsocie7y (approved):
GJ 👍

PR #519, opened 2023-11-08
Solution

D1st3f:


Wonsky1 (approved):

db/models.py
```diff
@@ -25,6 +27,11 @@ class Movie(models.Model):
     def __str__(self) -> str:
         return self.title
 
+    class Meta:
+        indexes = [
+            models.Index(fields=["title"]),
+        ]
```
fsocie7y, line 33: class Meta should be placed before methods. fix it everywhere

db/models.py
```diff
@@ -50,3 +57,54 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(on_delete=models.CASCADE, to="User")
```
fsocie7y, line 64: You should refer to AUTH_USER_MODEL. Import it from settings.py

services/movie_session.py
```diff
@@ -42,3 +40,12 @@ def update_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
+    return [
+        {"row": item["ticket__row"], "seat": item["ticket__seat"]}
+        for item in MovieSession.objects.filter(
+            ticket__movie_session=movie_session_id
+        ).values("ticket__row", "ticket__seat")
```
fsocie7y, line 50: It's fine, as for me, but you could also create additional variable to increase readability

services/user.py
```diff
@@ -0,0 +1,45 @@
...
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+
+
+def get_user(user_id: int) -> User:
```
fsocie7y, line 24: u can use `get_user_model` for type annotation too)

fsocie7y (changes requested):

fsocie7y (approved):
Well done ^_^

PR #523, opened 2023-11-08
Solution

MiskoRuslan:


Katherine-Greg (approved):

db/models.py
```diff
@@ -50,3 +52,57 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey("User", on_delete=models.CASCADE)
```
fsocie7y, line 59: You should refer to AUTH_USER_MODEL. Import it from settings.py

db/models.py
```diff
@@ -50,3 +52,57 @@ class MovieSession(models.Model):
...
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+    class Meta:
+        ordering = ["-created_at"]
```
fsocie7y, line 65: class Meta should be placed before methods. fix it everywhere

services/order.py
```diff
@@ -0,0 +1,31 @@
+from db.models import Ticket, Order, User, MovieSession
+from django.db import transaction
+from django.db.models.query import QuerySet
+
+
+def create_order(tickets: list[dict],
+                 username: str,
+                 date: str = None) -> None:
+    with transaction.atomic():
+        customer_user = User.objects.get(username=username)
```
fsocie7y, line 10: Use get_user_model() instead of User. Check out the [checklist](https://github.com/mate-academy/py-movie-moviesession-cinemahall/blob/master/checklist.md#code-efficiency). Fix it everywhere)

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+from django.contrib.auth import get_user_model
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
```
fsocie7y, line 11: please try to use one codestyle) 
image
or next
image

fsocie7y (changes requested):

services/order.py
fsocie7y: i will aprove, but fix it here also)

fsocie7y (approved):

PR #526, opened 2023-11-09
py-tickets-and-orders Solution

RVChornyy:


db/models.py
```diff
@@ -50,3 +57,56 @@ class MovieSession(models.Model):
...
+                "seat": f"seat number must be in available range: "
+                        f"(1, seats_in_row):"
+                        f" (1, {self.movie_session.cinema_hall.seats_in_row})"
+            })
+
+    def save(self,
+             force_insert: bool = False,
+             force_update: bool = False,
+             using: str = None,
+             update_fields: list = None) -> super:
```
RabenkoYevhenii, line 101: i`m not sure, but maybe bool is better here. "super" is not obvious

RabenkoYevhenii (approved):
NIce

db/models.py
```diff
@@ -50,3 +57,56 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
```
fsocie7y, line 64: You should refer to AUTH_USER_MODEL. Import it from settings.py

db/models.py
```diff
@@ -50,3 +57,56 @@ class MovieSession(models.Model):
...
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+    class Meta:
+        ordering = ["-created_at"]
```
fsocie7y, line 70: class Meta should be placed before methods.

fsocie7y: fix everywhere

fsocie7y (changes requested):

fsocie7y (approved):
GJ 🥇

PR #534, opened 2023-11-10
solution

StepanShevchuk:


db/models.py
```diff
@@ -50,3 +53,48 @@ class MovieSession(models.Model):
...
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
ArsenPidhoretskyi, line 70: In this case better to use str(self.created_at)

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
ArsenPidhoretskyi, line 10:
```suggestion
    title: Optional[str] = None
```
Because None by default

services/movie_session.py
```diff
@@ -42,3 +42,14 @@ def update_movie_session(
...
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    seats = []
+    for ticket in (
+            MovieSession.objects.get(
+            id=movie_session_id
+            ).ticket_set.all()):
+        seats.append({"row": ticket.row,
+                      "seat": ticket.seat})
```
ArsenPidhoretskyi, line 54: Simplify your code, and use filtering by Ticket.objects.filter() instead of MovieSession

services/order.py
```diff
@@ -0,0 +1,26 @@
...
+            Ticket.objects.create(row=ticket["row"],
+                                  seat=ticket["seat"],
+                                  movie_session_id=ticket["movie_session"],
+                                  order=new_order)
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
+    return Order.objects.all()
```
ArsenPidhoretskyi, line 26:
```suggestion
    orders = Order.objects.all()
    if username:
        orders = orders.filter(user__username=username)
    return orders
```
Better to write like this. It helps to support code in future

services/movie.py
ArsenPidhoretskyi: FIx in all places(only your code)

ArsenPidhoretskyi (changes requested):

services/order.py
```diff
@@ -0,0 +1,31 @@
...
+            new_order.created_at = date
+            new_order.save()
+        for ticket in tickets:
+            Ticket.objects.create(row=ticket["row"],
+                                  seat=ticket["seat"],
+                                  movie_session_id=ticket["movie_session"],
+                                  order=new_order)
+
+
+def get_orders(username: Optional[str] = None) -> QuerySet:
```
vasylhnatiuk, line 27:
```suggestion
def get_orders(username: Optional[str] = None) -> QuerySet[Order]:
```

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #539, opened 2023-11-13
Develop

Andru37:


services/movie.py
```diff
@@ -28,13 +31,16 @@ def create_movie(
...
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+
+    with transaction.atomic():
```
EdAlekseiev, line 35: Use `transaction.atomic` as decorator instead of context manager

services/movie_session.py
```diff
@@ -40,5 +41,21 @@ def update_movie_session(
...
+
+    for ticket in tickets:
+        taken_seats.append(
+            {
+                "row": ticket.row,
+                "seat": ticket.seat
+            }
+        )
+
+    return taken_seats
```
EdAlekseiev, line 57: Use method `value` of QuerySet here

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> None:
+
+    taken_seats = []
+    with transaction.atomic():
```
EdAlekseiev, line 16: Use` transaction.atomic` as decorator instead of context manager

EdAlekseiev (changes requested):

services/movie.py
Andru37: I didn't know there was such a decorator, now I know thanks

services/movie_session.py
Andru37: Thank you very much I didn't think of that before

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+from db.models import Order, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> None:
+
+    taken_seats = []
```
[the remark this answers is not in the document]
Andru37: Because I made a mistake and I really wanted to fix it

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+            ticket = Ticket.objects.create(
+                order=order,
+                movie_session_id=ticket_date["movie_session"],
+                row=ticket_date["row"],
+                seat=ticket_date["seat"],
+            )
+            # **ticket_date why this is not working, because movie_session
+            # in ticket_date and I need movie_session_id
+
+            taken_seats.append(ticket)
```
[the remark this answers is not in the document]
Andru37: Likewise

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+from django.db.models import QuerySet
+from django.db.models.functions import datetime
+
+from db.models import Order, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
```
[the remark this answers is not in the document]
Andru37: I can't do this  Pycharm crying

EdAlekseiev (approved):
Good job

PR #544, opened 2023-11-18
Solutions "Tickets and orders"

mtiunov:


services/user.py
```diff
@@ -0,0 +1,51 @@
...
+
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None,
```
YukihiroSM, line 11: If the default value is None - you should use Optional[str] or str | None typehint.

YukihiroSM: fix it in all code

YukihiroSM (changes requested):

NikOneZ1 (approved):
Well done!

PR #551, opened 2023-11-30
Solution

olgierrd:


olgierrd:
do not understand why checks are failing

olgierrd:
ok, added one migration and worked)

db/migrations/0003_alter_ticket_unique_together.py
```diff
@@ -0,0 +1,17 @@
...
+    dependencies = [
+        ('db', '0002_user_order_ticket_alter_movie_actors_and_more'),
+    ]
+
+    operations = [
+        migrations.AlterUniqueTogether(
+            name='ticket',
+            unique_together={('movie_session', 'row', 'seat')},
+        ),
+    ]
```
Dimosphen1, line 17: It is a good practice for Django projects to include one migration per app per pull request to not overload the `migrations` folder in the future. Add one migration to your pull request.

db/models.py
```diff
@@ -50,3 +59,73 @@ class MovieSession(models.Model):
...
+                    f"(1, seats_in_row): (1, {hall_seats})"
+                ]
+            })
+
+    def save(self,
+             force_insert: bool = False,
+             force_update: bool = False,
+             using: bool = None,
+             update_fields: bool = None
+             ) -> None:
```
Dimosphen1, line 117: When you do not have any arguments on the line with a closing parenthesis, your open parenthesis should not have any arguments on the same line as well. Add this change to your code style.

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
Dimosphen1, line 36: When you apply `transaction.atomic()` to the whole function body, it is a better choice to use it as a decorator instead of a context manager.

services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: id) -> list[dict]:
+    return list(
+        get_movie_session_by_id(movie_session_id)
+        .tickets.values("row", "seat")
+        .all()
```
Dimosphen1, line 51: There is no need to add `.all()` at the end of this expression.

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+
+from db.models import Order, User
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> Order:
+    with transaction.atomic():
```
Dimosphen1, line 12: The comment regatding `transaction.atomic()` relates here as well.

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+                row=ticket["row"],
+                seat=ticket["seat"],
+            )
+        order.save()
+        return order
+
+
+def get_orders(
+        username: str = None,
+) -> QuerySet:
```
Dimosphen1, line 30: You could specify a `QuerySet` of what model instances you return in the return type annotation. Make this change here and in other views where it is applicable.

Dimosphen1 (changes requested):

db/migrations/0003_alter_ticket_unique_together.py
olgierrd: It will not work otherwise, already tried, read at the top, before the last commit

vsmutok (approved):
Good Job 👍

PR #553, opened 2023-12-04
Solution

kravetb:


services/movie.py
```diff
@@ -2,11 +2,14 @@
 
 from db.models import Movie
 
+from django.db import transaction
+
 
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
```
EdAlekseiev, line 10: If argument can be `None` you should show it in type hint. Fix everywhere.
```suggestion
    genres_ids: list[int] | None = None,
    actors_ids: list[int] | None = None,
```

EdAlekseiev (changes requested):

PR #556, opened 2023-12-08
Solution

Molodoirockstar:


services/order.py
```diff
@@ -0,0 +1,46 @@
...
+            movie_session_id = ticket_data["movie_session"]
+
+            Ticket.objects.create(
+                order=order,
+                row=row,
+                seat=seat,
+                movie_session_id=movie_session_id
+            )
+
+    return order
```
vsmutok, line 37: When you apply transaction.atomic() to the whole function body, it is a better choice to use it as a decorator instead of a context manager.

vsmutok (approved):
Good Job!

PR #557, opened 2023-12-09
solution_py-tickets-and-orders_1

VladKrit-commits:


services/order.py
```diff
@@ -0,0 +1,32 @@
...
+    for ticket in tickets:
+        Ticket.objects.create(
+            row=ticket["row"],
+            seat=ticket["seat"],
+            movie_session_id=ticket["movie_session"],
+            order=order
+        )
+
+
+def get_orders(username: str = None) -> Order:
```
viktoria-rybenchuk, line 28: Does this function return `Order`?

viktoria-rybenchuk (changes requested):

services/movie.py
```diff
@@ -28,13 +34,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
Dimosphen1, line 37: In case you apply `transaction.atomic` to the whole function body, you could use `transaction.atomic` as a decorator for a function.

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@transaction.atomic()
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
+    order = Order.objects.create(
+        user=User.objects.get(
```
Dimosphen1, line 13: Use `get_user_model()` instead of `User` here and in other places where applicable as mentioned in the checklist.

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+    for ticket in tickets:
+        Ticket.objects.create(
+            row=ticket["row"],
+            seat=ticket["seat"],
+            movie_session_id=ticket["movie_session"],
+            order=order
+        )
+
+
+def get_orders(username: str = None) -> QuerySet:
```
Dimosphen1, line 29: You could additionally specify that a `QuerySet` of `Order` is returned in the return type annotation.

Polyakiv-Andrey (approved):
Good job!

PR #562, opened 2023-12-13
Solution

yevpy:


services/movie.py
```diff
@@ -1,11 +1,13 @@
...
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None,
+        title: str | None = None
```
YukihiroSM, line 10:
```suggestion
        genres_ids: list[int] | None = None,
        actors_ids: list[int] | None = None,
        title: str | None = None
```

YukihiroSM: Fix this everywhere.

services/movie_session.py
```diff
@@ -25,10 +25,10 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
...
 def update_movie_session(
-    session_id: int,
-    show_time: str = None,
-    movie_id: int = None,
-    cinema_hall_id: int = None,
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None,
 ) -> None:
```
YukihiroSM, line 32: same as above

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+from db.models import Order, Ticket
+from services.movie_session import get_taken_seats, get_movie_session_by_id
+
+
+@transaction.atomic
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
```
YukihiroSM, line 14: same as above

YukihiroSM (changes requested):

EdAlekseiev (approved):
Good job

PR #566, opened 2023-12-17
new_solution

Depaton:


db/models.py
```diff
@@ -1,4 +1,7 @@
+from django.contrib.auth.models import AbstractUser
```
Dimosphen1, line 1: Please do not delete the existing migration, create a new one if you need to.

db/models.py
```diff
@@ -50,3 +56,55 @@ class MovieSession(models.Model):
...
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True, )
```
Dimosphen1, line 66: There is no need to leave a trailing `, ` at the field definition.

models.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+class DjangoMigrations(models.Model):
+    app = models.CharField(max_length=255)
+    name = models.CharField(max_length=255)
+    applied = models.DateTimeField()
+
+    class Meta:
+        managed = False
+        db_table = "django_migrations"
```
Dimosphen1, line 184: There is no need for this file, create a regular migration in the `db` app to encounter all model changes and leave the default structure of the django DB initialization.

services/movie_session.py
```diff
@@ -42,3 +43,8 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
```
Dimosphen1, line 45: You should use the `get_movie_session_by_id` method to get the required `MovieSession` object.

services/movie_session.py
```diff
@@ -42,3 +43,8 @@ def update_movie_session(
...
+
+
+def get_taken_seats(movie_session_id: int) -> list:
+    tickets = Ticket.objects.filter(movie_session_id=movie_session_id)
```
Dimosphen1, line 49: It would be better to use `.values("row", "seat")` in the query to obtain relevant results without the need for list comprehension. Rewrite this method using this approach.

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@transaction.atomic
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> Order:
+    user = User.objects.get(username=username)
+    with transaction.atomic():
```
Dimosphen1, line 15: There is no need to have `transaction.atomic` decorator and context manager at the same time, leave only the decorator.

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+                seat=ticket["seat"],
+                movie_session=MovieSession.objects.get(
+                    id=ticket["movie_session"]
+                ),
+                order=order,
+            )
+    return order
+
+
+def get_orders(username: str = None) -> Order:
```
Dimosphen1, line 34: Check your type annotations in the code, you return not a single `Order` object, but a `QuerySet` of `Order` objects. Change your type annotation to encounter that.

Dimosphen1 (changes requested):

db/models.py
Dimosphen1: This comment is not implemented, return the first migration and make a second one with all required changes.

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+            seat=ticket["seat"],
+            movie_session=MovieSession.objects.get(
+                id=ticket["movie_session"]
+            ),
+            order=order,
+        )
+    return order
+
+
+def get_orders(username: str = None) -> QuerySet:
```
Dimosphen1, line 34: You could indicate that you return a `QuerySet` of `Order` objects in the return type annotation. Please, make changes to the return type annotation in all relevant places to encounter that.

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #573, opened 2024-01-01
completed the task

BGBroth:


settings.py
```diff
@@ -25,4 +25,8 @@
 
 INSTALLED_APPS = [
     "db",
```
Dimosphen1, line 27: Created apps should be placed after system modules in `INSTALLED_APPS` as a general practice.

services/order.py
```diff
@@ -0,0 +1,44 @@
...
+from db.models import Order, Ticket, User
+from datetime import datetime
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> Order:
+    with transaction.atomic():
```
Dimosphen1, line 13: You could use `transaction.atomic()` as a decorator instead of a context manager in case all function body should be treated as an atomic transaction.

services/movie.py
```diff
@@ -15,6 +17,9 @@ def get_movies(
     if actors_ids:
         queryset = queryset.filter(actors__id__in=actors_ids)
 
+    if title:
+        queryset = queryset.filter(Q(title__icontains=title))
```
Dimosphen1, line 21: It seems that there is no need for `Q` here.

Dimosphen1 (approved):
LGTM, well done! Several comments are left for consideration.

PR #579, opened 2024-01-06
Sol1

Davison322:


db/models.py
```diff
@@ -50,3 +57,65 @@ class MovieSession(models.Model):
...
+        ordering = ["-created_at"]
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession,
+                                      on_delete=models.CASCADE,
+                                      related_name="tickets")
+    order = models.ForeignKey(Order,
+                              on_delete=models.CASCADE,
+                              related_name="tickets")
```
vsmutok, line 85:
```suggestion
    movie_session = models.ForeignKey(
        MovieSession,
        on_delete=models.CASCADE,
        related_name="tickets"
    )
    order = models.ForeignKey(
        Order,
        on_delete=models.CASCADE,
        related_name="tickets"
    )
```

vsmutok: fix indentation

you can use `black` to do this automatically

services/movie.py
```diff
@@ -28,13 +32,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
vsmutok, line 35: I think better use transaction.atomic decorator

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+from django.db.models import QuerySet
+from db.models import Order, Ticket, User
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> Order:
+    with transaction.atomic():
```
vsmutok, line 12: I think better use transaction.atomic decorator

vsmutok (changes requested):

PR #580, opened 2024-01-10
Solution

aleksandra-prognimak:


db/models.py
```diff
@@ -50,3 +57,90 @@ class MovieSession(models.Model):
...
+        using: str = None,
+        update_fields: list = None
+    ) -> None:
+        self.full_clean()
+        super(Ticket, self).save(
+            force_insert=False,
+            force_update=False,
+            using=None,
+            update_fields=None
+        )
```
LiudmylaKulinchenko, line 135: You should call super with arguments, that will be passed in the function here. No need to set them to default values

db/models.py
```diff
@@ -50,3 +57,90 @@ class MovieSession(models.Model):
...
+
+    def save(
+        self,
+        force_insert: bool = False,
+        force_update: bool = False,
+        using: str = None,
+        update_fields: list = None
+    ) -> None:
+        self.full_clean()
+        super(Ticket, self).save(
```
LiudmylaKulinchenko, line 130: After the last updates, there is no need to pass Ticket, self to super(). It will take them automatically

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+from django.db.models import QuerySet
+from db.models import Order, Ticket, User, MovieSession
+
+
+def create_order(
+    tickets: list[dict],
+    username: str,
+    date: str = None
+) -> Order:
+    with transaction.atomic():
```
LiudmylaKulinchenko, line 11: u can use decorator instead of  context manager here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job ☀

PR #589, opened 2024-01-17
Solution

YegorT:


services/movie.py
```diff
@@ -1,11 +1,13 @@
...
+from django.db import transaction
+from django.db.models import QuerySet, Q
 
 from db.models import Movie
 
 
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
EdAlekseiev, line 10: If argument can be `None` you should show it in type hint. Fix everywhere
```suggestion
    genres_ids: list[int] | None = None,
    actors_ids: list[int] | None = None,
    title: str | None = None
```

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
EdAlekseiev, line 36: Use `transaction.atomic` as decorator instead of context manager

EdAlekseiev (changes requested):

PR #600, opened 2024-01-31
Develop

maxvell4:


db/migrations/0001_initial.py
```diff
@@ -1,58 +1,114 @@
...
+                ('seat', models.IntegerField()),
+                ('movie_session', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='db.moviesession')),
+                ('order', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='db.order')),
+            ],
+        ),
+        migrations.AddIndex(
+            model_name='movie',
+            index=models.Index(fields=['title'], name='db_movie_title_5d0841_idx'),
+        ),
+    ]
```
Dimosphen1, line 114: Do not modify the existing migration, add one new migration to the `db` app that will have all required changes to the models.

db/models.py
```diff
@@ -42,11 +51,102 @@ def __str__(self) -> str:
 class MovieSession(models.Model):
     show_time = models.DateTimeField()
     cinema_hall = models.ForeignKey(
-        to=CinemaHall, on_delete=models.CASCADE, related_name="movie_sessions"
+        to=CinemaHall, on_delete=models.CASCADE, related_name="cinema_hall"
     )
     movie = models.ForeignKey(
-        to=Movie, on_delete=models.CASCADE, related_name="movie_sessions"
+        to=Movie, on_delete=models.CASCADE, related_name="movie"
```
Dimosphen1, line 57: Do not change the `related_name` is was defined correctly.

db/models.py
```diff
@@ -42,11 +51,102 @@ def __str__(self) -> str:
...
+        on_delete=models.CASCADE,
+        null=False
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    constraints = [
+        UniqueConstraint(
+            fields=["row", "seat", "movie_session"],
+            name="unique_ticket")]
```
Dimosphen1, line 94: Place a closing bracket on a new line so that it would not have any arguments on the same line as well as the opening bracket.

db/models.py
```diff
@@ -42,11 +51,102 @@ def __str__(self) -> str:
...
+            seat=self.seat,
+            movie_session=self.movie_session
+        )
+
+        if existing_tickets.exists():
+            raise ValidationError("Not Unique Value")
+
+        if not (
+                self.row <= self.movie_session.cinema_hall.rows
+        ):
```
Dimosphen1, line 108: There is no need to complicate this check with `not`, you could omit it
```suggestion
        if self.row > self.movie_session.cinema_hall.rows:
```

Dimosphen1 (changes requested):

db/models.py
maxvell4: изображение

PR #601, opened 2024-01-31
Solution

nicksetrakov:


db/models.py
```diff
@@ -50,3 +59,72 @@ class MovieSession(models.Model):
...
+            raise ValidationError({
+                "row": f"row number must be in available range: "
+                       f"(1, rows): (1, {self.movie_session.cinema_hall.rows})"
+            })
+
+    def save(self,
+             force_insert: bool = False,
+             force_update: bool = False,
+             using: Any | None = None,
+             update_fields: Any | None = None) -> None:
```
LiudmylaKulinchenko, line 127: you can just set Any, no need to add `| None`

db/models.py
```diff
@@ -50,3 +59,72 @@ class MovieSession(models.Model):
...
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        User,
```
LiudmylaKulinchenko, line 71: Use get_user_model() instead of User. Check out the [checklist](https://github.com/mate-academy/py-movie-moviesession-cinemahall/blob/master/checklist.md#code-efficiency).

db/models.py
```diff
@@ -50,3 +59,72 @@ class MovieSession(models.Model):
...
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket"
+            )
+        ]
+
+    def __str__(self) -> str:
+        return f"{self.movie_session.movie.title} " \
+               f"{self.movie_session.show_time} " \
+               f"(row: {self.row}, seat: {self.seat})"
```
LiudmylaKulinchenko, line 108: Don't use ` \` as a linebreaker, create temporary variable instead or use parentheses

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```
LiudmylaKulinchenko, line 36: you can also use decorator here instead of context manager

services/order.py
```diff
@@ -0,0 +1,43 @@
...
+        order = Order.objects.create(user=user)
+
+        if date:
+            order.created_at = date
+        order.save()
+
+        for ticket in tickets:
+            row = ticket.get("row")
+            seat = ticket.get("seat")
+            movie_session = ticket.get("movie_session")
```
LiudmylaKulinchenko, line 26: Since this data is necessary to create a ticket, you can access a value directly using a key, without using get method

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(
+        user_id: int,
+        username: str | None = None,
+        password: str | None = None,
+        email: str | None = None,
+        first_name: str | None = None,
+        last_name: str | None = None
```
LiudmylaKulinchenko, line 35: You can just add None as defaul and do not specify None in annotations

LiudmylaKulinchenko (changes requested):
Please, make sure you fixed my comments throughout the code
Looks good overall!

Sofbog (approved):
LGTM

PR #604, opened 2024-02-09
Done

Moonralex:


db/migrations/0001_initial.py
```diff
@@ -40,19 +69,58 @@ class Migration(migrations.Migration):
...
+        migrations.AddField(
+            model_name='user',
+            name='user_permissions',
+            field=models.ManyToManyField(blank=True, help_text='Specific permissions for this user.', related_name='user_set', related_query_name='user', to='auth.Permission', verbose_name='user permissions'),
+        ),
+        migrations.AddConstraint(
+            model_name='ticket',
+            constraint=models.UniqueConstraint(fields=('row', 'seat', 'movie_session'), name='unique_row_seats_session'),
+        ),
     ]
```
Dimosphen1, line 126: Do not edit the migrations that are already created in the project and applied to the database. Create a new one instead and include it in the PR.

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    tickets = (
+        Ticket.objects
+        .filter(movie_session_id=movie_session_id)
+        .values_list("row", "seat")
+    )
```
Dimosphen1, line 52: There is no need to allocate a variable and after that perform a list comprehension, use `values` instead of `values_list` which includes keys by default.

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+
+from db.models import User, Order, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> Order:
+    with transaction.atomic():
```
Dimosphen1, line 12: In case you would like the whole body of the function to be treated as an atomic transaction, you could use `transaction.atomic` as a decorator.

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+            Ticket.objects.get_or_create(
+                movie_session_id=ticket["movie_session"],
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+            )
+    return order
+
+
+def get_orders(username: str = None) -> QuerySet:
```
Dimosphen1, line 28: Indicate a queryset of what instances you return in a return type annotation.

services/user.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = "",
+        last_name: str = ""
+) -> User:
+    with transaction.atomic():
```
Dimosphen1, line 18: There is no need to perform this operation with atomic since it includes only the creation of the single user.

services/user.py
```diff
@@ -0,0 +1,68 @@
...
+                validate_password(password)
+                user.set_password(raw_password=password)
+            except ValidationError as e:
+                raise e
+        if email:
+            try:
+                validate_email(email)
+                user.email = email
+            except ValidationError as e:
+                raise e
```
Dimosphen1, line 61: Instead of creating three try except blocks, you could leave one and perform all operations there, including placement of the if conditions and evaluation. Besides that, your error handling block does not differ between different try except clauses.

Dimosphen1 (changes requested):

PR #620, opened 2024-02-19
Solution2

artemgrishko:


services/movie.py
```diff
@@ -1,14 +1,19 @@
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
+    title: str = None,
```
VladKyianytsia, line 8: make type hints accurate everywhere. If it can be None by default, it means you need to add | None

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+
+from django.db import transaction
+
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket, User
+
+
+def create_order(tickets: list, username: str, date: datetime = None) -> None:
+    user = User.objects.get(username=username)
```
VladKyianytsia, line 11: Use get_user_model() instead of User, it is the best practice

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+    with transaction.atomic():
+        order = Order.objects.create(user=user)
+        if date:
+            order.created_at = date
+            order.save()
+
+        for ticket_info in tickets:
+            row = ticket_info["row"]
+            seat = ticket_info["seat"]
+            movie_session_id = ticket_info["movie_session"]
```
VladKyianytsia, line 21: cant it be done inside Ticket.objects.create()?

services/user.py
```diff
@@ -0,0 +1,54 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = User.objects.get(pk=user_id)
```
VladKyianytsia, line 37: You can use your get_user func here

PR #625, opened 2024-02-19
Solution

kstorozhenko:


db/models.py
```diff
@@ -50,3 +56,54 @@ class MovieSession(models.Model):
...
+        if not 1 <= self.row <= 10:
+            raise ValidationError(
+                {"row": "row number must be in available range:"
+                        " (1, rows): (1, 10)"}
+            )
+        if not 1 <= self.seat <= 12:
+            raise ValidationError(
+                {"seat": "seat number must be in available range:"
+                         " (1, seats_in_row): (1, 12)"}
+            )
```
aLEKS-e3, line 96: why did you hardcode the number of seats and rows here?

services/movie_session.py
```diff
@@ -24,6 +24,15 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
     return MovieSession.objects.get(id=movie_session_id)
 
 
+def get_taken_seats(movie_session_id: int) -> list:
+    movie_session = MovieSession.objects.get(id=movie_session_id)
```
aLEKS-e3, line 28: you can use `get_movie_session_by_id` func here

services/movie_session.py
```diff
@@ -24,6 +24,15 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
...
+    tickets = Ticket.objects.filter(movie_session=movie_session)
+    taken_seats = [
+        {"row": ticket.row, "seat": ticket.seat} for ticket in tickets
+    ]
+    return taken_seats
```
aLEKS-e3, line 33: I think you can return the result directly without creating a `taken_seats` variable

services/order.py
```diff
@@ -0,0 +1,45 @@
...
+from db.models import Order, Ticket, User
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime | None = None,
+) -> None:
+    with transaction.atomic():
+        user = User.objects.get(username=username)
```
aLEKS-e3, line 18: read the checklist, it's a bad practice to use `User` directly

services/order.py
```diff
@@ -0,0 +1,45 @@
...
+        order = Order.objects.create(
+            user=user)
+        if date:
+            order.created_at = date
+
+        order.save()
```
aLEKS-e3, line 24: I think it's better to move this save statement to the `if` block so as not to make a query when not needed

services/order.py
```diff
@@ -0,0 +1,45 @@
...
+            row = ticket["row"]
+            seat = ticket["seat"]
+            movie_session_id = ticket["movie_session"]
+
+            Ticket.objects.create(
+                order=order,
+                row=row,
+                seat=seat,
+                movie_session_id=movie_session_id
+            )
```
aLEKS-e3, line 36: why didn't you pass those values directly?

services/order.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username is None:
+        return Order.objects.all()
+
+    user_model = get_user_model()
+    user = user_model.objects.get(username=username)
+    return Order.objects.filter(user=user)
```
aLEKS-e3, line 45: you can filter user like this - `user__username=username`

aLEKS-e3 (changes requested):

PR #627, opened 2024-02-19
Sol

Dobryk12:


Dobryk12:
> sorry for this one)

No problem, a good score on the tech check and we're even.

PR #632, opened 2024-02-19
Solution

Serejka21:


db/models.py
```diff
@@ -50,3 +55,77 @@ class MovieSession(models.Model):
...
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_tickets_value"
+            )
+        ]
+
+    def __str__(self) -> str:
+        movie = self.movie_session.movie.title
+        session_time = self.movie_session.show_time
+        return (f"{movie} {session_time} "
```
aLEKS-e3, line 100: why don't you just use `self.movie_session.movie`? (you have a `str` method)

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+                 date: datetime | None = None) -> None:
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
+
+        order = Order.objects.create(user=user)
+
+        if date:
+            order.created_at = date
+
+        order.save()
```
aLEKS-e3, line 20: i may be wrong, but isn't it better to place the `save` statement in the `if` block so as not to make a query when not needed?

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session_id=ticket.get("movie_session"),
+                order=order,
+                row=ticket.get("row"),
+                seat=ticket.get("seat"),
+            )
+
+
+def get_orders(username: str | None = None) -> list[str]:
```
aLEKS-e3, line 31: are you getting a list of str here?

services/user.py
```diff
@@ -0,0 +1,52 @@
...
+from db.models import User
+
+
+def create_user(username: str,
+                password: str,
+                email: str | None = None,
+                first_name: str | None = None,
+                last_name: str | None = None) -> None:
+
+    user = User.objects.create_user(username=username,
```
aLEKS-e3, line 12: better use `get_user_model` here)

aLEKS-e3 (changes requested):

PR #634, opened 2024-02-20
Task Solution

shaiduchyk:


services/user.py
```diff
@@ -0,0 +1,53 @@
...
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(pk=user_id)
```
tskopenko: use `get_user_model()` instead of `User`

tskopenko (approved):
change it, all else Looks good

PR #635, opened 2024-02-20
Solution

tskopenko:


db/models.py
```diff
@@ -1,5 +1,9 @@
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
 from django.db import models
 
+import settings
```
msymonovych, line 5: i think it's better to use `from django.conf import settings`

msymonovych (approved):

PR #642, opened 2024-02-25
Solution

AntonSkyrda:


services/user.py
```diff
@@ -0,0 +1,47 @@
+from db.models import User
+
+
+def create_user(username: str,
+                password: str,
+                email: str = "",
+                first_name: str = "",
+                last_name: str = "",
+                ) -> None:
+    user = User.objects.create_user(username=username,
```
Serejka21, line 10: Here is good practice to use get_user_model() method. Change where u need it

Serejka21 (approved):

PR #643, opened 2024-02-25
Solution

EugeneProtsenko:


services/order.py
```diff
@@ -0,0 +1,32 @@
+from django.contrib.auth import get_user_model
+from django.db import transaction
+
+from db.models import Ticket, Order, MovieSession
+
+
+def create_order(tickets: list,
```
vasylhnatiuk, line 7:
```suggestion
def create_order(tickets: list[dict],

```

vasylhnatiuk (changes requested):
delete db.sqlite3 from pull request

EdAlekseiev (approved):

PR #655, opened 2024-03-03
Solution

achernyshovv:


services/order.py
```diff
@@ -0,0 +1,40 @@
...
+
+from db.models import Order, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: date | None = None
+) -> None:
+    with transaction.atomic():
```
NEkropo1, line 15: since entire function is under transaction, better use a simple decorator.

NEkropo1 (changes requested):

PR #657, opened 2024-03-04
Solution

vsupruniuk:


services/user.py
```diff
@@ -0,0 +1,54 @@
...
+
+def update_user(
+    user_id: str,
+    username: str = None,
+    password: str = None,
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None,
+) -> None:
+    user = User.objects.get(id=user_id)
```
LiudmylaKulinchenko, line 37: Use get_user_model instead of User

LiudmylaKulinchenko: fix everywhere

db/models.py
```diff
@@ -50,3 +57,75 @@ class MovieSession(models.Model):
...
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        to=User,
```
LiudmylaKulinchenko, line 69: you should refer to AUTH_USER_MODEL

db/models.py
```diff
@@ -50,3 +57,75 @@ class MovieSession(models.Model):
...
+
+
+class Ticket(models.Model):
+    row = models.IntegerField()
+    seat = models.IntegerField()
+    movie_session = (models.ForeignKey
+                     (to=MovieSession,
+                      on_delete=models.CASCADE,
+                      related_name="tickets")
+                     )
```
LiudmylaKulinchenko, line 89: fix indentation here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🚀

PR #661, opened 2024-03-06
solution

oleksandr-lysanets:


services/order.py
```diff
@@ -0,0 +1,35 @@
...
+
+from db.models import Order, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
+    with transaction.atomic():
```
EdAlekseiev, line 13: Use transaction.atomic as decorator

EdAlekseiev (changes requested):

NikOneZ1 (approved):
Goode job!

PR #662, opened 2024-03-08
created new models, modified existing, created services

RomanHlodann:


services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
vsmutok, line 36: In case all body of a method should be treated as an atomic transaction, it is better to use `transaction.atomic` as a decorator

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+
+from db.models import Order, Ticket
+
+
+def create_order(
+    tickets: list[dict],
+    username: str,
+    date: str = None
+) -> Order:
+    with transaction.atomic():
```
vsmutok, line 15: In case all body of a method should be treated as an atomic transaction, it is better to use `transaction.atomic` as a decorator

vsmutok (approved):

PR #669, opened 2024-03-19
'Solution'

yura-push:


db/migrations/0004_ticket_ticket_unique_ticket.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

services/movie.py
```diff
@@ -28,13 +35,13 @@ def create_movie(
...
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
+    with transaction.atomic():
```
Dimosphen1, line 38: In case all body of a method should be treated as an atomic transaction, it is better to use `transaction.atomic` as a decorator

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+
+from db.models import Order, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> Order:
+    with transaction.atomic():
```
Dimosphen1, line 13: In case all body of a method should be treated as an atomic transaction, it is better to use `transaction.atomic` as a decorator

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #671, opened 2024-03-20
Develop

twonumberfortyfives:


services/movie.py
```diff
@@ -6,20 +7,24 @@
...
     if actors_ids:
         queryset = queryset.filter(actors__id__in=actors_ids)
 
+    if title:
+        queryset = queryset.filter(title__icontains=title)
+
     return queryset
 
 
 def get_movie_by_id(movie_id: int) -> Movie:
```
pavlopro, line 25: If you decide to modify it, it would be great if you could also update the return annotation to include None as a possible return value.

services/movie.py
```diff
@@ -6,20 +7,24 @@
...
-    return Movie.objects.get(id=movie_id)
+    if movie_id:
+        return Movie.objects.get(id=movie_id)
 
 
 def create_movie(
```
pavlopro, line 30: Please use decorator transaction.atomic for better readability

services/movie_session.py
```diff
@@ -42,3 +42,17 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    taken_seats = []
+    if movie_session_id:
+        movie_session = MovieSession.objects.get(pk=movie_session_id)
```
pavlopro, line 50: Please use `get_movie_session_by_id ` method

services/order.py
```diff
@@ -0,0 +1,36 @@
+from datetime import datetime
+
+from django.db import transaction
+
+from db.models import Order, Ticket, User
+
+
+def create_order(
```
pavlopro, line 8: The same with decorator here as well

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+                order.created_at = date
+                order.save()
+            for ticket_data in tickets:
+                ticket = Ticket.objects.create(
+                    movie_session_id=ticket_data["movie_session"],
+                    row=ticket_data["row"],
+                    seat=ticket_data["seat"],
+                    order=order,
+                )
+                ticket.save()
```
pavlopro, line 27: No need to keep it

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+from django.db import transaction
+
+from db.models import Order, Ticket, User
+
+
+def create_order(
+        tickets: list[Ticket],
+        username: str,
+        date: datetime = None,
+) -> Order:
```
pavlopro, line 12: Please adjust the return annotation to include None as well.

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+                ticket.save()
+            return order
+
+
+def get_orders(username: str = None) -> list[Order]:
+    if username:
+        return Order.objects.filter(
+            user__username=username
+        ).order_by("-created_at")
+    return Order.objects.all().order_by("-created_at")
```
pavlopro, line 36: No need to specify -created_at as the default ordering is already defined in the Meta class

services/user.py
```diff
@@ -0,0 +1,53 @@
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> User:
```
pavlopro, line 10: Please adjust the return annotation here and in the methods below as well.

pavlopro (changes requested):

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+from db.models import Order, Ticket, User
+
+
+@transaction.atomic()
+def create_order(
+        tickets: list[Ticket],
+        username: str,
+        date: datetime = None,
+) -> Order | None:
+    with transaction.atomic():
```
pavlopro, line 14: Remove with transaction.atomic() as you already use decorator

pavlopro (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #672, opened 2024-03-20
Implement py-tickets-and-orders

anrozhna:


services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+
+from db.models import Order, User, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
+    with transaction.atomic():
```
Dimosphen1, line 13: In case all body of a method should be treated as an atomic transaction, it is better to use `transaction.atomic` as a decorator

Dimosphen1 (approved):
LGTM, well done! Several comments were left for consideration

PR #673, opened 2024-03-21
Solution

KwintDimon:


db/models.py
```diff
@@ -50,3 +57,73 @@ class MovieSession(models.Model):
...
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        formatted_date = self.created_at.strftime("%Y-%m-%d %H:%M:%S")
+        return formatted_date
```
Dimosphen1, line 75: Do not create `formatted_date` variable, return the value directly in the same places in the future

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #676, opened 2024-03-23
Solution

IjdmLoverI:


services/movie.py
```diff
@@ -28,13 +32,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
Dimosphen1, line 35: In case all body of a method should be treated as an atomic transaction, it is better to use `transaction.atomic` as a decorator

services/movie_session.py
```diff
@@ -42,3 +42,17 @@ def update_movie_session(
...
+        id=movie_session_id
+    ).values("tickets__row", "tickets__seat"))
+    result = []
+    for ticket in tickets:
+        result.append({
+            "row": ticket["tickets__row"],
+            "seat": ticket["tickets__seat"]
+        })
+
+    return result
```
Dimosphen1, line 58: You could build a query and avoid iteration by starting from the `Ticket` model, e.g. `Ticket.objects.filter(movie_session_id=...`. Rewrite this method taking this into account

services/order.py
```diff
@@ -0,0 +1,30 @@
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order
+from db.models import Ticket
+from db.models import User
+
+
+def create_order(tickets: list[dict], username: str, date: str = None) -> None:
+    with transaction.atomic():
```
Dimosphen1, line 10: Use `transaction.atomic()` as a decorator here

services/order.py
```diff
@@ -0,0 +1,30 @@
...
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+
+def get_orders(username: str = None) -> QuerySet:
```
Dimosphen1, line 26: Indicate in a return type annotation a `QuerySet` of what instance types you return

services/order.py
```diff
@@ -0,0 +1,30 @@
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order
+from db.models import Ticket
+from db.models import User
```
Dimosphen1, line 6: Import these models in one line, do not duplicate `from db.models` for several times

Dimosphen1 (changes requested):

PR #682, opened 2024-03-28
Solution

mish-sk:


services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
EdAlekseiev, line 10: If argument can be None, you should show it in type hint

services/order.py
```diff
@@ -0,0 +1,39 @@
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket, User, MovieSession
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
```
EdAlekseiev, line 10: If argument can be None, you should show it in type hint

services/user.py
```diff
@@ -0,0 +1,49 @@
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
```
EdAlekseiev, line 9: If argument can be None, you should show it in type hint

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+    return User.objects.get(id=user_id)
+
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
```
EdAlekseiev, line 35: If argument can be None, you should show it in type hint

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #683, opened 2024-03-29
Solution

vburkalo:


services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```
vsmutok, line 36: In case all body of a method should be treated as an atomic transaction, it is better to use `transaction.atomic` as a decorator

vsmutok (changes requested):

EdAlekseiev (approved):
Good job

PR #686, opened 2024-04-02
Solution

Bujhm99:


db/models.py
```diff
@@ -41,12 +50,60 @@ def __str__(self) -> str:
 
 class MovieSession(models.Model):
     show_time = models.DateTimeField()
-    cinema_hall = models.ForeignKey(
-        to=CinemaHall, on_delete=models.CASCADE, related_name="movie_sessions"
-    )
-    movie = models.ForeignKey(
-        to=Movie, on_delete=models.CASCADE, related_name="movie_sessions"
-    )
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
```
Oleksl888, line 53: why remove related_names here?

db/models.py
```diff
@@ -41,12 +50,60 @@ def __str__(self) -> str:
...
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return (str(self.movie) + " "
```
Oleksl888, line 57: Why use str concatenation instead of f-strings?

db/models.py
```diff
@@ -41,12 +50,60 @@ def __str__(self) -> str:
...
+
+    class Meta:
+        ordering = ["-created_at"]
+
+
+class Ticket(models.Model):
+    row = models.IntegerField()
+    seat = models.IntegerField()
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
```
Oleksl888, line 77: what about related names for these fields?

db/models.py
```diff
@@ -41,12 +50,60 @@ def __str__(self) -> str:
...
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        # unique_together = ["seat", "trip"]
+        constraints = [
+            models.UniqueConstraint(fields=["seat", "row", "movie_session"],
+                                    name="unick_ticket")
```
Oleksl888, line 104: constraint name should be more descriptive, it needs to tell exactly what it does e.g. `row_seat_per_session_uniq`

services/cinema_hall.py
```diff
@@ -7,9 +7,16 @@ def get_cinema_halls() -> QuerySet:
     return CinemaHall.objects.all()
 
 
-def create_cinema_hall(
-    hall_name: str, hall_rows: int, hall_seats_in_row: int
-) -> CinemaHall:
-    return CinemaHall.objects.create(
-        name=hall_name, rows=hall_rows, seats_in_row=hall_seats_in_row
-    )
+def create_cinema_hall(hall_name: str,
```
Oleksl888, line 10: There was no requirement to reformat this

services/order.py
```diff
@@ -0,0 +1,25 @@
...
+        for ticket in tickets:
+            Ticket.objects.create(row=ticket["row"], seat=ticket["seat"],
+                                  movie_session_id=ticket["movie_session"],
+                                  order_id=order.pk)
+
+
+def get_orders(username: str = None) -> QuerySet:
+    orders = Order.objects.all()
+    if username:
+        orders = orders.filter(user_id=User.objects.get(username=username).pk)
```
Oleksl888, line 24: Use `get_user_model` to get user instead

services/user.py
```diff
@@ -0,0 +1,60 @@
...
+    User.objects.create_user(username=username,
+                             email=(email or ""),
+                             first_name=(first_name or ""),
+                             last_name=(last_name or ""),
+                             password=password)
+
+
+def get_user(user_id: int) -> User | None:
+    try:
+        man = User.objects.get(id=user_id)
```
Oleksl888, line 33: Using `get_user_model` to get user for project is a recommended way

services/user.py
```diff
@@ -0,0 +1,60 @@
...
+    except ObjectDoesNotExist:
+        return None
+    return man
+
+
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: models.EmailField = None,
```
Oleksl888, line 42: Email validators are usually used when working with form data. You don't really have to include it here but if you do, why not include validation in other fields then?

Oleksl888 (changes requested):

db/models.py
```diff
@@ -13,14 +17,19 @@ class Actor(models.Model):
     last_name = models.CharField(max_length=255)
 
     def __str__(self) -> str:
-        return f"{self.first_name} {self.last_name}"
+        return self.first_name + " " + self.last_name
```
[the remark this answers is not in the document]
Bujhm99: In some doc to exsercise was writen not to use f strings in such construction like it.

services/user.py
Oleksl888: I will rephrase myself. I don't think validation is necessary here - it usually is done on the model level. The same as you have done in your clean method on Ticket. 
And on the other hand I was wondering why did you choose to add extra validation only for email, but not to other fields and other models

Oleksl888: you don't have to do any extra validation other then specified in the task requirements btw

Bujhm99:
I used validation, because when i try to update user's mail with wrong email format (for exemple fkdjgGmail.com) it updates wthout any problems. Added some validation for name (in order it can't be not alphabet), but not username

Oleksl888:
> I used validation, because when i try to update user's mail with wrong email format (for exemple fkdjgGmail.com) it updates wthout any problems. Added some validation for name (in order it can't be not alphabet), but not username

Yeah it’s probably because build in Django user model doesn’t have email validation from the box

db/models.py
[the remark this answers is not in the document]
Oleksl888: This was in boiler plate code for the task so I guess you don't need to change it unless it is specified in task.

db/models.py
```diff
@@ -41,12 +50,69 @@ def __str__(self) -> str:
...
+                {"row": f"row number must be in available range: "
+                        f"(1, rows): (1, {quant_of_row})"}
+            )
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        # unique_together = ["seat", "trip"]
```
Oleksl888, line 110: what is this comment for?

services/cinema_hall.py
```diff
@@ -7,9 +7,16 @@ def get_cinema_halls() -> QuerySet:
...
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> CinemaHall:
+    return CinemaHall.objects.create(name=hall_name,
+                                     rows=hall_rows,
+                                     seats_in_row=hall_seats_in_row)
+
+
+def get_cinema_hall_by_id(cinema_hall_id: int) -> CinemaHall | None:
```
Oleksl888, line 18: There was no requirement to make this but even if you had to, why would you return None if cinema hall doesn't exist? Shouldn't warning be more appropriate here?

services/order.py
```diff
@@ -0,0 +1,28 @@
+from datetime import datetime
```
Oleksl888, line 1: missing blank line here

services/order.py
```diff
@@ -0,0 +1,28 @@
...
+from django.db.transaction import atomic
+from django.db.models import QuerySet
+from django.contrib.auth import get_user_model
+
+from db.models import Order, Ticket
+
+User = get_user_model()
```
Oleksl888, line 8: better use a different name here since it shadows django built-in `User` class

services/user.py
```diff
@@ -0,0 +1,68 @@
+from django.db import models
+from django.core.exceptions import ValidationError
```
Oleksl888, line 2: order of import here + need to add separators between them

services/user.py
```diff
@@ -0,0 +1,68 @@
...
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.validators import validate_email
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
+
+
+def email_validator(email: str) -> None:
```
Oleksl888, line 10: As I mentioned earlier, validation would be more appropriate inside the model. You can redefine the fields for User and overwrite clean() method there

settings.py
```diff
@@ -24,5 +24,7 @@
 USE_TZ = False
 
 INSTALLED_APPS = [
-    "db",
+    "db", "django.contrib.auth", "django.contrib.contenttypes"
```
Oleksl888, line 27: it is a convention to put every app name on a new line

Oleksl888 (changes requested):

services/cinema_hall.py
Bujhm99: I don’t remember why I did it, but what I did it for was not included in the solution, so it was redundant, I deleted it.

db/models.py
Bujhm99: Same as below.

Bujhm99:
61dbGcJXvCL _AC_UF894,1000_QL80_
Big bright picture for bookmark

db/models.py
```diff
@@ -1,4 +1,8 @@
 from django.db import models
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
+
+import settings
```
Oleksl888, line 5: `from django.conf import settings` this is the correct way of importing setting in a project

db/models.py
```diff
@@ -41,12 +50,71 @@ def __str__(self) -> str:
...
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=["seat", "row", "movie_session"],
+                                    name="unique_ticket_row_seat_per_session")
+        ]
+
+
+class User(AbstractUser):
+
+    def save(self, *args, **kwargs) -> None:
```
Oleksl888, line 118: what does this accomplish if you haven't specified clean method here?

services/order.py
```diff
@@ -0,0 +1,31 @@
...
+            Ticket.objects.create(row=ticket["row"], seat=ticket["seat"],
+                                  movie_session_id=ticket["movie_session"],
+                                  order_id=order.pk)
+
+
+def get_orders(username: str = None) -> QuerySet:
+    orders = Order.objects.all()
+    if username:
+        orders = orders.filter(
+            user_id=User_model.objects.get(username=username).pk
```
Oleksl888, line 29: here you can use lookup so you optimize your query `orders = orders.filter(user__username=username)`

services/order.py
```diff
@@ -0,0 +1,31 @@
...
+        if date:
+            order.created_at = date
+        order.save()
+        for ticket in tickets:
+            Ticket.objects.create(row=ticket["row"], seat=ticket["seat"],
+                                  movie_session_id=ticket["movie_session"],
+                                  order_id=order.pk)
+
+
+def get_orders(username: str = None) -> QuerySet:
```
Oleksl888, line 25: for optional parameters like these is better to use `str | None or Optonal[str]` for typing

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+                                   first_name=(first_name or ""),
+                                   last_name=(last_name or ""),
+                                   password=password)
+
+
+def get_user(user_id: int) -> User | None:
+    try:
+        man = User_model.objects.get(id=user_id)
+    except ObjectDoesNotExist:
+        return None
```
Oleksl888, line 27: No need to explicitly return None here - let the exception happen and it may be try-excepted in place where this service is used

Oleksl888 (changes requested):

Bujhm99:
image


class User(AbstractUser):

    def save(self, *args, **kwargs) -> None:
@[Oleksl888](https://github.com/Oleksl888) Oleksl888 [3 days ago](https://github.com/mate-academy/py-tickets-and-orders/pull/686#discussion_r1571898744)
what does this accomplish if you haven't specified clean method here? 

It runs full_clean(), which runs validators for email

Oleksl888 (approved):
Well done!

PR #688, opened 2024-04-02
Implemented Tickets and orders

esemesenko:


services/movie.py
```diff
@@ -1,11 +1,12 @@
 from django.db.models import QuerySet
-
 from db.models import Movie
+from django.db import transaction
 
 
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: list[str] = None,
```
EdAlekseiev, line 9: If argument can be `None` you should show in in type hint. Fix everywhere, please
```suggestion
    title: list[str] | None = None,
```

services/movie.py
```diff
@@ -28,13 +32,14 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```
EdAlekseiev, line 35: Use transaction.atomic as decorator

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+from datetime import datetime
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> Order:
+
+    with transaction.atomic():
```
EdAlekseiev, line 14: Use transaction.atomic as decorator

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #689, opened 2024-04-03
'Solution'

KravchenkoSerhiii:


db/migrations/0003_alter_order_created_at.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

db/models.py
```diff
@@ -50,3 +58,65 @@ class MovieSession(models.Model):
...
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return self.created_at.strftime("%Y-%m-%d %H:%M:%S")
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
```
Dimosphen1, line 79: Add `related_name` to `ForeignKey` fields, please

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```

services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    movie_session = MovieSession.objects.get(id=movie_session_id)
+    return [
+        {"row": ticket.row, "seat": ticket.seat}
+        for ticket in movie_session.ticket_set.all()
+    ]
```
Dimosphen1, line 52: Retrieve a list of required dicts in one line using Django ORM. You [could use](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#values) `.values` for it

services/order.py
```diff
@@ -0,0 +1,31 @@
+from django.contrib.auth import get_user_model
+from django.db import transaction
+from db.models import User, Order, MovieSession, Ticket
+
+
+def create_order(tickets: list[dict], username: str, date: str = None) -> None:
+    user = get_user_model().objects.get_or_create(username=username)
+
+    with transaction.atomic():
```
Dimosphen1, line 9: You could place user inside of an atomic transaction and apply it as a decorator afterwards

services/user.py
```diff
@@ -0,0 +1,58 @@
...
+    if first_name:
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
+
+
+def get_user(user_id: int) -> get_user_model():
```
Dimosphen1, line 28: Return back `User` in the return type annotation, do not use function calls for type annotations

Dimosphen1 (changes requested):

services/movie.py
```diff
@@ -15,26 +17,31 @@ def get_movies(
...
```
EdAlekseiev, line 37:
```suggestion
```

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
EdAlekseiev, line 10: If argument can be `None`, you should show it in type hint. Fix everywhere
```suggestion
    title: str | None = None,
```

EdAlekseiev (changes requested):

vsmutok (approved):
Great!

PR #695, opened 2024-04-12
Solution

RomanNest:


EdAlekseiev (changes requested):
Fix flake8, please

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
```suggestion
    title: str | None = None,
```

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+
+from db.models import Order, Ticket, MovieSession
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str | None = None,
+) -> Order:
+    with transaction.atomic():
```
EdAlekseiev, line 13: Use transaction.atomic as decorator

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #697, opened 2024-04-13
Solution

Boyok1337:


db/models.py
```diff
@@ -50,3 +59,56 @@ class MovieSession(models.Model):
...
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
```
LiudmylaKulinchenko, line 78: add related names

db/models.py
```diff
@@ -50,3 +59,56 @@ class MovieSession(models.Model):
...
+
+    def __str__(self) -> str:
+        return (f"{self.movie_session.movie.title} "
+                f"{self.movie_session.show_time} "
+                f"(row: {self.row}, "
+                f"seat: {self.seat})")
+
+    def clean(self) -> None:
+        valid_row = self.movie_session.cinema_hall.rows
+        if self.row not in range(1, valid_row + 1):
```
LiudmylaKulinchenko, line 96: why dont you just check `if not (1 <= self.row <= self.movie_session.cinema_hall.rows)` instead of creating range?

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+                 username: str,
+                 date: datetime = None,
+                 ) -> None:
+    user = get_user_model().objects.get(username=username)
+    new_order = Order.objects.create(user_id=user.id)
+
+    if date:
+        new_order.created_at = date
+
+    new_order.save()
```
LiudmylaKulinchenko, line 20: You can perform save inside `if date` block

LiudmylaKulinchenko (changes requested):

db/models.py
Boyok1337: i think there is the same speed and memory, so just choose this way to release this  func like this, but now i know that range is not good idea

db/models.py
```diff
@@ -50,3 +59,60 @@ class MovieSession(models.Model):
...
+    order = models.ForeignKey(Order,
+                              on_delete=models.CASCADE,
+                              related_name="ticket_order")
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=["movie_session", "row", "seat"],
+                                    name="unique_ticket")
```
Oleksl888, line 89: is better to have meaningful name for contstraint such as row_seat_per_session_uniq

db/models.py
```diff
@@ -50,3 +59,60 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
```
Oleksl888, line 61: I think we should follow this datetime format `2022-03-15 15:59:50` please check if you comply everywhere

Oleksl888 (changes requested):

db/models.py
```diff
@@ -49,4 +58,62 @@ class MovieSession(models.Model):
...
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession,
+                                      on_delete=models.CASCADE,
+                                      related_name="session_tickets")
```
Oleksl888, line 80: just tickets would be fine :)

db/models.py
```diff
@@ -49,4 +58,62 @@ class MovieSession(models.Model):
...
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(settings.AUTH_USER_MODEL,
+                             on_delete=models.CASCADE)
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
Oleksl888, line 74: here should be also formatted

Oleksl888 (approved):
Good job!

db/models.py
Boyok1337: `    movie_session = models.ForeignKey(MovieSession,
                                      on_delete=models.CASCADE,
                                      related_name="session_tickets")
    order = models.ForeignKey(Order,
                              on_delete=models.CASCADE,
                              related_name="order_tickets")`
                              
i can`t name it tickets, cause i need to have diff between this 2 backward relation, how i can distinguish if they will have same names?

Oleksl888: Simple because they will exist in different models. One will be MovieSession.tickets another Order.tickets

PR #701, opened 2024-04-18
Tickets and orders

Ananiev-Vitalii:


services/movie.py
```diff
@@ -1,33 +1,34 @@
...
+from django.db import transaction
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None,
+        title: str = None,
```
EdAlekseiev, line 9: if argument can be None, you should show it in type hint. Fix everywhere
```suggestion
        genres_ids: list[int] | None = None,
        actors_ids: list[int] | None = None,
        title: str | None = None,
```

EdAlekseiev (changes requested):

PR #703, opened 2024-04-18
Solution

mwellick:


services/movie.py
```diff
@@ -1,11 +1,12 @@
...
+from django.db import transaction
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None,
+        title: str = None
```
EdAlekseiev, line 9: Why did you move arguments to right on 4 spaces?
```suggestion
    genres_ids: list[int] = None,
    actors_ids: list[int] = None,
    title: str = None
```

services/movie.py
```diff
@@ -23,18 +27,19 @@ def get_movie_by_id(movie_id: int) -> Movie:
...
 
 def create_movie(
-    movie_title: str,
-    movie_description: str,
-    genres_ids: list = None,
-    actors_ids: list = None,
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list = None,
+        actors_ids: list = None,
```
EdAlekseiev, line 33: The same comment as above

services/movie_session.py
```diff
@@ -1,10 +1,10 @@
 from django.db.models import QuerySet
 
-from db.models import MovieSession
+from db.models import MovieSession, Ticket
 
 
 def create_movie_session(
-    movie_show_time: str, movie_id: int, cinema_hall_id: int
+        movie_show_time: str, movie_id: int, cinema_hall_id: int
```
EdAlekseiev, line 7: The same comment as above

services/movie_session.py
```diff
@@ -25,10 +25,10 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
...
 
 def update_movie_session(
-    session_id: int,
-    show_time: str = None,
-    movie_id: int = None,
-    cinema_hall_id: int = None,
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None,
```
EdAlekseiev, line 31: The same comemnt as above

services/order.py
```diff
@@ -0,0 +1,33 @@
+from django.db import transaction
+from db.models import Order, Ticket
+from django.contrib.auth import get_user_model
+
+
+def create_order(tickets: list[dict],
+                 username: str,
+                 date: str | None = None) -> Order:
+    with transaction.atomic():
```
EdAlekseiev, line 9: Use transaction.atomic as decorator instance of context manager. It makes your code more readable

EdAlekseiev (changes requested):

services/movie.py
mwellick: I use ctrl + alt + L  shortcut  to format code and I thought it's good

mwellick: I always used 4 spaces.It's 1st time when I had to make it 2

EdAlekseiev (approved):

PR #708, opened 2024-04-24
Solution

vitalinamal:


services/order.py
```diff
@@ -0,0 +1,40 @@
...
+from db.models import Order, Ticket, MovieSession, User
+
+
+def create_order(
+        tickets: list[dict],
+        username: str, date:
+        datetime = None
+) -> None:
+    with transaction.atomic():
+        user = User.objects.get(username=username)
```
YuriiKindrat, line 15: it is beter to use `get_user_model()` here instead of User

services/user.py
```diff
@@ -0,0 +1,47 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = User.objects.get(pk=user_id)
```
YuriiKindrat, line 34: same is here for user

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -1,4 +1,8 @@
+from django.contrib.auth.models import AbstractUser
 from django.db import models
+from django.core.exceptions import ValidationError
+
+import settings
```
Oleksl888, line 5: importing settings from `django.conf` is a prefereable way

db/models.py
```diff
@@ -50,3 +54,72 @@ class MovieSession(models.Model):
...
+        related_name="tickets"
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique ticket"
```
Oleksl888, line 92: it is common to set constraint names very verbose so they describe what exactly they are doing. Also no spaces in names. Here is example how it may look like : `row_seat_session_uniq`

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
```
Oleksl888, line 8: It is a good practice to use Optional type hint for optional params

services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict] | str:
```
Oleksl888, line 47: check return type here

services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
...
+    tickets = Ticket.objects.filter(movie_session_id=movie_session_id)
```
Oleksl888, line 48: Not a mistake, but why not call function `get_movie_session_by_id` defined above?

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+                pk=ticket_data["movie_session"]
+            )
+            ticket_instances.append(Ticket(
+                row=ticket_data["row"],
+                seat=ticket_data["seat"],
+                movie_session=movie_session,
+                order=order
+            ))
+
+        Ticket.objects.bulk_create(ticket_instances)
```
Oleksl888, line 34: This solution look a bit complicated, you definitely benefit from using bulk_create but also you have to create movie_session object on each iteration + bulk_create doesn't trigger full_clean method on create. I'd say more optimal solution here would be something like this:
```
        for ticket_data in tickets:
            1. prepare ticket data (add movie_session_id - just an id no need to create instance, add order)
            2. Ticket.objects.create(**ticket_data)

File: services/order.py [hunk 9]
```diff
@@ -0,0 +1,41 @@
...
+
+
+def get_orders(username: str = None) -> Order | QuerySet:
```
Oleksl888, line 37: return type will always be Queryset here

Oleksl888 (changes requested):

services/movie_session.py
```diff
@@ -42,3 +44,13 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
```
Oleksl888, line 49: What I meant was this:
```
movie_session = get_movie_session_by_id(movie_session_id)
    return [{"row": ticket.row, "seat": ticket.seat} for ticket in movie_session.ticket_set.all()]

Oleksl888 (approved):
Well done! Just fix that `get_taken_seats` function

services/movie_session.py
vitalinamal: Understood
Done!)

PR #709, opened 2024-04-25
Cinema DB solution

panicua:


db/models.py
```diff
@@ -1,5 +1,9 @@
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
 from django.db import models
 
+import settings
```
Oleksl888, line 5: importing settings from `django.conf` is a prefereable way

db/models.py
```diff
@@ -50,3 +54,63 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.CASCADE
```
Oleksl888, line 62: related_name here?

services/movie.py
```diff
@@ -1,18 +1,23 @@
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+        genres_ids: list[int] = None,
```
Oleksl888, line 8: use Optional type hint here and similar places

services/movie.py
```diff
@@ -1,18 +1,23 @@
...
+        actors_ids: list[int] = None,
+        title: str = None
 ) -> QuerySet:
     queryset = Movie.objects.all()
 
-    if genres_ids:
+    if title is not None:
+        queryset = queryset.filter(title__contains=title)
+
+    if genres_ids is not None:
```
Oleksl888, line 17: if genres_ids is perfectly fine here :)

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
+        movie = Movie.objects.create(
+            title=movie_title,
+            description=movie_description,
+        )
+        if genres_ids is not None:
```
Oleksl888, line 41: if genres_ids is a better choice

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    movie_session = MovieSession.objects.get(pk=movie_session_id)
```
Oleksl888, line 48: why not use func ` get_movie_session_by_id` defined just above?

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
...
+    tickets = movie_session.tickets.all()
```
Oleksl888, line 49: you don't really need to store it in variable

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+                movie_session_id=ticket["movie_session"]
+            )
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
+    orders = Order.objects.all()
+
+    if username is not None:
+        user_id = get_user_id_by_username(username)
+        orders = orders.filter(user_id=user_id)
```
Oleksl888, line 34: just use lookup here by `user__username`

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = get_user_model().objects.create_user(
+        username=username, password=password
+    )
+    if email is not None:
```
Oleksl888, line 16: just write `if email`  in cases like this unless you really really need to check for None value

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = get_user(user_id)
+
+    if username is not None:
```
Oleksl888, line 39: same as above..

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def get_user(user_id: int) -> User:
+    return get_user_model().objects.get(pk=user_id)
+
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
```
Oleksl888, line 32: update type hints here too - and other places

Oleksl888 (changes requested):

services/movie.py
panicua: > if genres_ids is perfectly fine here :)

what if ```genres_ids = []``` instead of ```None```, we have to lookup at it to check what we ment, I belive it gives more readablility when we specify ```is not None```

Oleksl888: empty list evaluates as false no?

panicua: > empty list evaluates as false no?

probably right here, I don't remember the case but I once asked GPT about checking for None if value = None and he said that condition ```is not None``` were prefereable there.

services/movie.py
```diff
@@ -1,14 +1,21 @@
...
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+        genres_ids: Optional = None,
```
Oleksl888, line 10: It is also a good practice to specify what is expected in optional: `Optional[list] = None` for example. 
Also these lines are overindented

services/movie.py
Oleksl888: Well GPT changes his mind pretty easy :)
`is not None` assumes we are checking only if param is actually None but other falsy values may get through

services/order.py
```diff
@@ -0,0 +1,37 @@
+from typing import Optional
+
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Ticket, Order
+from services.user import get_user_id_by_username
```
Oleksl888, line 7: this creates a dependency of this module on other service. Lets keep them isolated

Oleksl888 (changes requested):

services/order.py
panicua: I'll just remove that function since it's short + used only in one function after refactoring

services/movie.py
panicua: NVIDIA_Share_zSop26mCtA
I don't understand where are they overintended, they are not repeatable in my code nomore

db/models.py
```diff
@@ -50,3 +54,65 @@ class MovieSession(models.Model):
...
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return self.created_at.strftime("%Y-%m-%d %H:%M:%S")
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        MovieSession, on_delete=models.DO_NOTHING, related_name="tickets")
```
YuriiKindrat, line 76: bad formatting here (

db/models.py
```diff
@@ -50,3 +54,65 @@ class MovieSession(models.Model):
...
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_row_seat_movie_session"
+            )
+        ]
```
YuriiKindrat, line 110: it is better to place class meta before other methods for better readability

YuriiKindrat (changes requested):

db/models.py
panicua: I see, didn't pay enough attention

db/models.py
panicua: will know now, I only heard that it's good to put it after attributes, didn't know that's better to be the first after them.

services/movie.py
```diff
@@ -1,14 +1,21 @@
...
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+        genres_ids: Optional[list[int]] = None,
```
Oleksl888, line 10: still overindented ( must be 4 spaces )

Oleksl888 (approved):
Approved. But please fix indentation in last comment

services/movie.py
panicua: > still overindented ( must be 4 spaces )

when I press enter to move attributes to new line it drops it to that place, I never did it manually, ctrl+alt+L also does same

PR #711, opened 2024-04-25
Solution

SashaKisliy:


.flake8
Oleksl888, line 1: you don't need to push these changes to remote

db/models.py
```diff
@@ -50,3 +59,61 @@ class MovieSession(models.Model):
...
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(fields=[
+                "movie_session",
+                "row",
+                "seat", ],
+                name="unique_movie_session")
```
Oleksl888, line 93: Better use more descriptive contraint name e.g. `uniq_session_row_seat` so it's obvious what constraint does from the first sight

services/movie.py
```diff
@@ -6,7 +7,11 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
Oleksl888, line 10: Use `Optional` type hint for optional params

services/order.py
```diff
@@ -0,0 +1,37 @@
+from datetime import datetime
+
+from django.contrib.auth import get_user_model
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket
+
+
+def create_order(tickets: list[Ticket],
```
Oleksl888, line 10: wrong type here

services/user.py
```diff
@@ -0,0 +1,56 @@
+from django.contrib.auth import get_user_model
+
+from db.models import User
+
+
+def create_user(username: str,
+                password: str,
+                email: str = None,
+                first_name: str = None,
```
Oleksl888, line 9: Use `Optional` here too

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+    user.save()
+
+    return user
+
+
+def get_user(user_id: int) -> User | str:
+    try:
+        return get_user_model().objects.get(pk=user_id)
+    except User.DoesNotExist:
+        return f"{user_id} does not exist"
```
Oleksl888, line 30: returning string here is not the best choice, with this implementation it may lead to errors in other places. So the best would be to exit the program, or just let exception happen and then try-catch it wherever you are calling this function

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> User | str:
+    try:
+        user = get_user_model().objects.get(pk=user_id)
+    except User.DoesNotExist:
+        return f"{user_id} does not exist"
```
Oleksl888, line 42: same as above

Oleksl888 (changes requested):

db/models.py
```diff
@@ -50,3 +59,61 @@ class MovieSession(models.Model):
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(settings.AUTH_USER_MODEL,
+                             on_delete=models.CASCADE,
+                             related_name="orders")
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+    class Meta:
+        ordering = ("-created_at",)
```
YuriiKindrat, line 74: it is better to place class meta before other methods

YuriiKindrat (approved):
GJ

PR #716, opened 2024-04-26
Solution 1

semermaks:


db/models.py
```diff
@@ -36,17 +39,71 @@ def capacity(self) -> int:
...
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = ForeignKey("User", on_delete=models.CASCADE)
```
YuriiKindrat, line 66: it is better to use AUTH_USER_MODEL instead of user, also what about related_name ?

db/models.py
```diff
@@ -36,17 +39,71 @@ def capacity(self) -> int:
...
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        "MovieSession", on_delete=models.CASCADE
+    )
+    order = models.ForeignKey("Order", on_delete=models.CASCADE)
```
YuriiKindrat, line 79: same here

db/models.py
```diff
@@ -36,17 +39,71 @@ def capacity(self) -> int:
...
+    def __str__(self) -> str:
+        return (f"{self.movie_session} "
+                f"(row: {self.row}, seat: {self.seat})")
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(fields=["movie_session",
+                                     "row", "seat"],
+                             name="unique_ticket")
+        ]
```
YuriiKindrat, line 109: it is better to place class Meta before all methods, for better readability

services/order.py
```diff
@@ -0,0 +1,25 @@
+from django.db import transaction
+from db.models import Order, Ticket, User
+
+
+@transaction.atomic
+def create_order(tickets: list[Ticket],
+                 username: str, date: str = None) -> None:
+    user = User.objects.get(username=username)
```
YuriiKindrat, line 8: it is better to use get_user_model() method here instead of User

services/order.py
```diff
@@ -0,0 +1,25 @@
...
+    if date:
+        order.created_at = date
+        order.save()
+    for ticket in tickets:
+        movie_session_id = ticket["movie_session"]
+        Ticket.objects.create(
+            order=order,
+            movie_session_id=movie_session_id,
+            row=ticket["row"],
+            seat=ticket["seat"])
```
YuriiKindrat, line 19: bad formatting here (

services/user.py
```diff
@@ -0,0 +1,38 @@
...
+        user.username = username
+    if password:
+        user.set_password(password)
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
```
YuriiKindrat, line 38: same is here, use get_user_model() method here

YuriiKindrat (changes requested):

PR #717, opened 2024-04-26
Solution

y-kondrashova:


db/models.py
```diff
@@ -50,3 +59,54 @@ class MovieSession(models.Model):
...
+    def __str__(self) -> str:
+        return self.created_at.strftime("%Y-%m-%d %H:%M:%S")
+
+    class Meta:
+        ordering = ["-created_at"]
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
```
YuriiKindrat, line 79: what about related names here ?

services/order.py
```diff
@@ -0,0 +1,35 @@
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import User, Order, Ticket
+
+
+@transaction.atomic
+def create_order(tickets: list, username: str, date: str = None) -> Order:
+    try:
+        user = User.objects.get(username=username)
```
YuriiKindrat, line 10: it is better to use get_user_model() method instead of User here

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+                movie_session_id=ticket_data["movie_session"]
+            )
+
+        return order
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
+    else:
```
YuriiKindrat, line 34: why do u need else here ?

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+from db.models import User
+
+
+def create_user(username: str,
+                password: str,
+                email: str = None,
+                first_name: str = "",
+                last_name: str = "") -> User:
+    user_model = get_user_model()
+    user = user_model.objects.create_user(
```
YuriiKindrat, line 12: no need to create variable for user model here

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> None:
+    try:
+        user = User.objects.get(id=user_id)
+    except User.DoesNotExist:
+        return None
```
YuriiKindrat, line 35: use method get_user_model here

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #720, opened 2024-04-26
Solution

SerhiiMar:


services/user.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None,
+) -> User:
+    user = User.objects.create_user(
```
YuriiKindrat, line 11: it is better to use get_user_model here, fix this everywhere

db/models.py
```diff
@@ -50,3 +60,74 @@ class MovieSession(models.Model):
...
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["movie_session", "row", "seat"],
+                name="unique_movie_session_row_seats"
+            ),
+        ]
```
YuriiKindrat, line 122: it is better to place class meta before other methods

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -1,4 +1,9 @@
 from django.db import models
+from django.db.models import UniqueConstraint
+from django.core.exceptions import ValidationError
+from django.contrib.auth.models import AbstractUser
+
+import settings
```
Oleksl888, line 6: import from 'django.conf' is a preferable way

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
Oleksl888, line 10: Using `Optional` type hint is a good practice for optional params - check in all functions too

services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return list(
+        MovieSession.objects.get(
```
Oleksl888, line 49: not a mistake but you could reuse ` get_movie_session_by_id` function defined above

services/user.py
```diff
@@ -0,0 +1,50 @@
...
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+
+    return user
+
+
+def get_user(user_id: int) -> User:
+    return get_user_model().objects.get(id=user_id)
```
Oleksl888, line 28: `pk` is a preferable way to query primary key in ORM

services/user.py
```diff
@@ -0,0 +1,50 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = get_user_model().objects.get(id=user_id)
```
Oleksl888, line 39: you can reuse `get_user` func here

Oleksl888 (changes requested):

db/models.py
```diff
@@ -50,3 +59,74 @@ class MovieSession(models.Model):
...
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
LiudmylaKulinchenko, line 73: Don't overuse f-strings, u had this in checklists for previous tasks. Use str() instead

db/models.py
LiudmylaKulinchenko: You didn't fix that

LiudmylaKulinchenko (changes requested):
Looks good overall, a few moments missed

services/user.py
```diff
@@ -0,0 +1,52 @@
...
+def create_user(
+        username: str,
+        password: str,
+        email: Optional[str] = None,
+        first_name: Optional[str] = None,
+        last_name: Optional[str] = None
+) -> User:
+    user = get_user_model().objects.create_user(
+        username=username,
+        email=email,
```
Oleksl888, line 17: why pass email field here since it's optional?

Oleksl888 (approved):
Well done! But please check last comment

services/user.py
SerhiiMar: builtin `create_user` function has optional email argument, so i decided that we don't need extra check for email

PR #722, opened 2024-04-26
Done

DenZatolokin:


services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```
EdAlekseiev, line 36: Use transaction.atomic as decorator instead of context manager. It makes your code more readable

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+
+from db.models import Order, Ticket, User
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime.datetime = None
+) -> None:
+    with transaction.atomic():
```

EdAlekseiev (changes requested):

services/user.py
```diff
@@ -0,0 +1,53 @@
...
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
+
+
+def get_user(
+        user_id: int
+) -> User:
```
vsmutok, line 29:
```suggestion
def get_user(user_id: int) -> User:
```

vsmutok (approved):
GJ!

PR #723, opened 2024-04-26
Solution

ayatsuta:


db/models.py
```diff
@@ -26,6 +28,56 @@ def __str__(self) -> str:
...
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(to="MovieSession",
+                                      on_delete=models.CASCADE)
```
vsmutok, line 44: bad formatting here

services/user.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None
+                ) -> None:
+    user = User.objects.get(pk=user_id)
```
vsmutok, line 34: use get_user_model instead of User everywhere

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```
vsmutok, line 36: In case all body of a method should be treated as an atomic transaction, it is better to use transaction.atomic as a decorator

vsmutok (changes requested):

services/order.py
```diff
@@ -0,0 +1,28 @@
+from django.db import transaction
+
+from db.models import Order, User, Ticket
+
+
+def create_order(tickets: list,
+                 username: str,
+                 date: str = None
+                 ) -> None:
+    with transaction.atomic():
```
EdAlekseiev, line 10: Use `transaction.atomic` as a decorator instead of a context manager. This makes your code more readable

EdAlekseiev (changes requested):

services/order.py
ayatsuta: I did used:)

ayatsuta: oh in order also

EdAlekseiev (approved):

PR #725, opened 2024-04-28
solution

CharmNiovi:


db/models.py
```diff
@@ -50,3 +62,62 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        get_user_model(), on_delete=models.CASCADE, related_name="users"
+    )
```
YuriiKindrat, line 71: are u sure about related name naming ?

services/user.py
```diff
@@ -0,0 +1,38 @@
...
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = get_user(user_id)
+    user.username = username if username else user.username
+    user.set_password(password) if password else None
```
YuriiKindrat, line 34: it will raise type error here in case of None

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #727, opened 2024-04-29
Solution

Ruxxell54:


services/order.py
```diff
@@ -0,0 +1,36 @@
...
+            order.created_at = date
+            order.save()
+
+    for ticket_data in tickets:
+        Ticket.objects.create(
+            row=ticket_data["row"],
+            seat=ticket_data["seat"],
+            movie_session_id=ticket_data["movie_session"],
+            order=order
+        )
```
YuriiKindrat, line 27: so why do u use transaction here, if tickets creation is out of transaction scope ?

YuriiKindrat (changes requested):

PR #728, opened 2024-04-29
Solution

messedxup:


db/models.py
```diff
@@ -42,11 +47,70 @@ def __str__(self) -> str:
...
+        ordering = ("-created_at",)
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(to=MovieSession,
+                                      on_delete=models.CASCADE)
+    order = models.ForeignKey(to=Order, on_delete=models.CASCADE)
```
YuriiKindrat, line 78: what about related names here ?

db/models.py
```diff
@@ -42,11 +47,70 @@ def __str__(self) -> str:
...
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(fields=["seat",
+                                     "row",
+                                     "movie_session"],
+                             name="unique_ticket_seat_row_movie_session"),
+        ]
```
YuriiKindrat, line 107: it is better to place meta class before all methods

services/order.py
```diff
@@ -0,0 +1,28 @@
+from django.db import transaction
+
+from db.models import Ticket, Order, User
+
+
+def create_order(tickets: list[dict],
+                 username: str,
+                 date: str = None) -> None:
+    with transaction.atomic():
+        user = User.objects.get(username=username)
```
YuriiKindrat, line 10: it is better to use get_user_model() method instead of User
fix this everywhere

services/order.py
```diff
@@ -0,0 +1,28 @@
...
+        order = Order.objects.create(user=user)
+        if date:
+            order.created_at = date
```
YuriiKindrat, line 13: why do u need this, created_at field will be filled automaticaly

services/user.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def create_user(username: str,
+                password: str,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> None:
+    create = User.objects.create_user(username=username,
+                                      password=password,
+                                      )
```
YuriiKindrat, line 11: bad formatting here

services/user.py
```diff
@@ -0,0 +1,42 @@
...
+    return User.objects.get(pk=user_id)
+
+
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> None:
+    user = User.objects.get(pk=user_id)
```
YuriiKindrat, line 31: u can use get user method here

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #729, opened 2024-04-29
Develop

VitalyBashkiser:


db/models.py
```diff
@@ -50,3 +55,72 @@ class MovieSession(models.Model):
...
+                "row",
+                "seat", ],
+                name="unique_session_row_seat")
+        ]
+
+    def __str__(self) -> str:
+        return (f"{self.movie_session.movie.title}"
+                f" {self.movie_session.show_time}"
+                f" (row: {self.row}, seat: {self.seat})"
+                )
```
YuriiKindrat, line 102: bad formatting here

services/movie_session.py
```diff
@@ -42,3 +44,12 @@ def update_movie_session(
...
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> List[Dict[str, int]]:
+    taken_seats = Ticket.objects.filter(
+        movie_session_id=movie_session_id
+    ).values(
+        "row", "seat"
+    )
+    return [{"row": seat["row"], "seat": seat["seat"]} for seat in taken_seats]
```
YuriiKindrat, line 55: could u just make like this here 
list(taken_seats), why do u need dict comprehension here ?

services/order.py
```diff
@@ -0,0 +1,41 @@
+from typing import List, Optional, Dict
+
+from django.db import transaction
+
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket
+
+from services.user import get_user_model
```
YuriiKindrat, line 9: did u create some custom method ?
u could use build in method 
from django.contrib.auth import get_user_model

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -1,4 +1,9 @@
 from django.db import models
+from django.core.exceptions import ValidationError
+from django.db.models import UniqueConstraint
+from django.contrib.auth.models import AbstractUser
+
+import settings
```
Oleksl888, line 6: import settings from `django.conf` is a preferable way

db/models.py
```diff
@@ -50,3 +55,71 @@ class MovieSession(models.Model):
...
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
Oleksl888, line 72: using f-string is possible here but calling `str()` would be a preferable way

db/models.py
```diff
@@ -50,3 +55,71 @@ class MovieSession(models.Model):
...
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(fields=[
+                "movie_session",
+                "row",
+                "seat", ],
```
Oleksl888, line 94: check formatting

services/movie_session.py
```diff
@@ -42,3 +44,12 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> List[Dict[str, int]]:
+    taken_seats = Ticket.objects.filter(
```
Oleksl888, line 50: no need to store it in variable - you can return result in one line

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+
+from django.db import transaction
+
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket
+
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
```
Oleksl888, line 11: what are you using this for?

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+    if date:
+        new_order.created_at = date
+        new_order.save()
+
+    for ticket in tickets:
+        Ticket.objects.create(
+            movie_session_id=ticket["movie_session"],
+            order_id=new_order.id,
+            seat=ticket["seat"],
+            row=ticket["row"])
```
Oleksl888, line 32: bad formatting

Oleksl888 (changes requested):

db/models.py
Oleksl888: not fixed

db/models.py
```diff
@@ -50,3 +55,71 @@ class MovieSession(models.Model):
...
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(fields=[
+                "movie_session",
+                "row",
+                "seat"],
```
Oleksl888, line 94: still not fixed

services/user.py
```diff
@@ -0,0 +1,52 @@
...
+
+def update_user(
+    user_id: Optional[int],
+    username: Optional[str] = None,
+    password: Optional[str] = None,
+    email: Optional[str] = None,
+    first_name: Optional[str] = None,
+    last_name: Optional[str] = None,
+) -> None:
+    user = get_user(user_id)
```
Oleksl888, line 40: if user does not exist then an exception will be triggered so basically here you will always have a user instance no need to use `if user` below (would be good for another use case but not here)

services/user.py
```diff
@@ -0,0 +1,52 @@
...
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
+
+
+def get_user(user_id: int) -> Optional[User]:
+    return User.objects.get(id=user_id)
```
Oleksl888, line 29: use `pk` instead

Oleksl888 (changes requested):

PR #731, opened 2024-05-01
Implement Tickets and orders task

bwaah1:


db/models.py
```diff
@@ -1,5 +1,13 @@
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
 from django.db import models
 
+import settings
```
Oleksl888, line 5: import settings from `django.conf` is a preferable way

db/models.py
```diff
@@ -42,11 +50,75 @@ def __str__(self) -> str:
...
+        to=settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
Oleksl888, line 79: Here `str()` method is preferable. f-string would be more appropriate if you have a string and you want to insert a variable in it

db/models.py
```diff
@@ -42,11 +50,75 @@ def __str__(self) -> str:
...
+                    f"seat number must be in available range: "
+                    f"(1, seats_in_row): "
+                    f"(1, {self.movie_session.cinema_hall.seats_in_row})"]}
+            )
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+    class Meta:
```
Oleksl888, line 119: class Meta as a convention is defined right after fields definition and above other methods

db/models.py
```diff
@@ -42,11 +50,75 @@ def __str__(self) -> str:
...
+        constraints = [
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"], name="unique_ticket"
```
Oleksl888, line 122: contstraint name should be more verbose - include all fields in name so its self explanatory

services/movie.py
```diff
@@ -6,9 +7,13 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
Oleksl888, line 10: use `Optional` type hint for optional params - check in other functions too

services/movie_session.py
```diff
@@ -42,3 +42,16 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    result = []
```
Oleksl888, line 48: this can be done in one line. You can use .values() method on tickets here

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+        user = get_user_model().objects.get(username=username)
+        new_order = Order.objects.create(user=user)
+
+        if date:
+            new_order.created_at = date
+            new_order.save()
+
+        for ticket_data in tickets:
+            movie_session_id = ticket_data.pop("movie_session")
+            movie_session = MovieSession.objects.get(pk=movie_session_id)
```
Oleksl888, line 25: no need for this - you can pass movie_session_id as is

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+            new_ticket = Ticket(
```
Oleksl888, line 26: why not use Ticket.objects.create here?

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+                movie_session=movie_session,
+                **ticket_data
+            )
+            new_ticket.save()
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        user = get_user_model().objects.get(username=username)
+        return Order.objects.filter(user=user)
```
Oleksl888, line 37: you can use lookup here instead of getting a user instance

services/user.py
```diff
@@ -0,0 +1,51 @@
+from django.contrib.auth import get_user_model
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = "",
```
Oleksl888, line 9: Mark Optional params with `Optional` type hint and set default value to None. Then instead of passing empty values on create use checks to fill out the data that actually exists. This would be preferable to writing empty values to database

Oleksl888 (changes requested):

services/user.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    return user
+
+
+def get_user(user_id: int) -> Optional[User]:
+    try:
+        return get_user_model().objects.get(pk=user_id)
+    except User.DoesNotExist:
+        return None
```
Oleksl888, line 31: So this is bad design - imagine you call your function on line 42 and user does not exist. The program will continue working and try to access `.username` or other fields on `None` object which will lead to errors.
Be careful with exceptions - when to use them and when not to, and also how to correctly process them.

Oleksl888 (changes requested):

PR #732, opened 2024-05-01
Solution

PozniakVA:


services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
EdAlekseiev, line 36: Use `transaction.atomic` as decorator instead of context manager. It makes your code more readable.

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+
+from db.models import Ticket, Order, User, MovieSession
+
+
+def create_order(
+        tickets: list[Ticket],
+        username: str,
+        date: datetime = None
+) -> None:
+    with transaction.atomic():
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #733, opened 2024-05-02
Solution1

yossosh:


db/models.py
```diff
@@ -1,4 +1,8 @@
 from django.db import models
+from django.core.exceptions import ValidationError
+from django.contrib.auth.models import AbstractUser
+
+import settings
```
Oleksl888, line 5: import from `django.conf` is preferable way

db/models.py
```diff
@@ -50,3 +57,73 @@ class MovieSession(models.Model):
...
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+
+    class Meta:
+        ordering = ("-created_at",)
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
Oleksl888, line 74: Here using str() method is preferable since you return only one value that you don't change

db/models.py
```diff
@@ -50,3 +57,73 @@ class MovieSession(models.Model):
...
+        related_name="tickets"
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["movie_session", "row", "seat"],
+                name="unique_ticket"
```
Oleksl888, line 95: contsraint name must be verbose and self-explanatory - you should mention all fields that form a constraint

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Ticket, Order
+
+
+@transaction.atomic
+def create_order(tickets: list[dict], username: str, date: str = None) -> None:
+    with transaction.atomic():
+        user_id = get_user_model().objects.get(username=username).id
```
Oleksl888, line 13: no need to store it in variable - since you are not going to reuse it

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+
+from django.contrib.auth import get_user_model
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Ticket, Order
+
+
+@transaction.atomic
+def create_order(tickets: list[dict], username: str, date: str = None) -> None:
```
Oleksl888, line 11: Better split this params on separate lines for readability + date missing Optional type hint

Oleksl888 (changes requested):

PR #734, opened 2024-05-03
Tickets and orders

Ananiev-Vitalii:


services/user.py
```diff
@@ -0,0 +1,45 @@
...
+def create_user(
+        username: str, password: str, email: str | None = None,
+        first_name: str | None = None, last_name: str | None = None
+) -> User:
+    user = User.objects.create_user(
+        username=username,
+        password=password,
+        email=email if email is not None else "",
+        first_name=first_name if first_name is not None else "",
+        last_name=last_name if last_name is not None else ""
```
vsmutok, line 16:
```suggestion
        email=email or "",
        first_name=first_name or "",
        last_name=last_name or ""
```
better use or in this case

services/user.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+    return user
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
```
vsmutok, line 22: it is better to use get_user_model() method instead of User
fix this everywhere

services/user.py
```diff
@@ -0,0 +1,45 @@
...
+@transaction.atomic
+def update_user(
+        user_id: int,
+        username: str | None = None,
+        password: str | None = None,
+        email: str | None = None,
+        first_name: str | None = None,
+        last_name: str | None = None
+) -> None:
+    user = User.objects.get(id=user_id)
```
vsmutok, line 34: u can use get user method here

vsmutok (changes requested):

services/user.py
Ananiev-Vitalii: @vsmutok User = get_user_model() But that's what I use?

PR #735, opened 2024-05-04
Tickets and orders

Ananiev-Vitalii:


services/movie.py
```diff
@@ -1,33 +1,34 @@
...
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+        genres_ids: list[int] | None = None,
+        actors_ids: list[int] | None = None,
+        title: str | None = None,
 ) -> QuerySet:
```
EdAlekseiev, line 10: You should specify inner type of `QuerySet`
```suggestion
) -> QuerySet[Movie]:
```

services/movie.py
```diff
@@ -1,33 +1,34 @@
...
-    movie_title: str,
-    movie_description: str,
-    genres_ids: list = None,
-    actors_ids: list = None,
-) -> Movie:
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list | None = None,
+        actors_ids: list | None = None,
+) -> QuerySet:
```
EdAlekseiev, line 31: Invalid type hint for return value. It returns Movie
```suggestion
) -> QuerySet[Movie]:
```

services/movie_session.py
```diff
@@ -25,10 +25,10 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
...
 
 def update_movie_session(
-    session_id: int,
-    show_time: str = None,
-    movie_id: int = None,
-    cinema_hall_id: int = None,
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None,
```
EdAlekseiev, line 31: if argument can be None, you should show in type hint
```suggestion
        show_time: str | None = None,
        movie_id: int | None = None,
        cinema_hall_id: int | None = None,
```

EdAlekseiev (changes requested):

PR #736, opened 2024-05-04
Solution

MykhailoIvchenko:


services/movie.py
```diff
@@ -6,9 +7,13 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
EdAlekseiev, line 10: If argument can be `None`, you should show it in type hint
```suggestion
    title: str | None = None,
```

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```
EdAlekseiev, line 36: Use `transaction.atomic` as decorator instead of context manager. It makes your code more readable

services/order.py
```diff
@@ -0,0 +1,46 @@
+from django.contrib.auth import get_user_model
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Ticket, Order, MovieSession
+
+
+def create_order(tickets: list[dict], username: str, date: str = None) -> None:
+    with transaction.atomic():
```
EdAlekseiev, line 9: The same comment as above

EdAlekseiev (changes requested):

services/movie.py
MykhailoIvchenko: Fixed. You always write about it and I always forget. Not good, not good at all, as Jack Reacher says))

services/movie.py
MykhailoIvchenko: Fixed

services/order.py
MykhailoIvchenko: Fixed

MykhailoIvchenko:
I admire, how fast you always react and check home tasks) Thanks a lot)

services/order.py
```diff
@@ -0,0 +1,47 @@
...
+        order.save()
+
+    movie_session_ids = set(ticket_data["movie_session"]
+                            for ticket_data in tickets)
+
+    movie_sessions = {
+        movie_session.id: movie_session
+        for movie_session
+        in MovieSession.objects.filter(id__in=movie_session_ids)
+    }
```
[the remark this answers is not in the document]
MykhailoIvchenko: I've written the new code but commented it because when I use only id the test fails with an error:
django.db.utils.IntegrityError: The row in table 'db_ticket' with primary key '1' has an invalid foreign key: db_ticket.movie_session_id contains a value '1' that does not have a corresponding value in db_moviesession.id.

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
...
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    tickets = Ticket.objects.filter(movie_session_id=movie_session_id)
+
+    taken_seats = [{"row": ticket.row, "seat": ticket.seat}
+                   for ticket in tickets]
+
+    return taken_seats
```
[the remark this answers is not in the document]
MykhailoIvchenko: Fixed

services/order.py
```diff
@@ -35,6 +35,16 @@ def create_order(tickets: list[dict],
         for ticket_data in tickets
     ]
 
+    # ticket_instances = [
```
vasylhnatiuk, line 38: delete comment

vasylhnatiuk (changes requested):

services/order.py
MykhailoIvchenko: Done

db/models.py
```diff
@@ -50,3 +60,63 @@ class MovieSession(models.Model):
...
+        to=Order, on_delete=models.CASCADE, related_name="tickets"
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    def __str__(self) -> str:
+        formatted_date = (
+            self.movie_session.show_time.strftime("%Y-%m-%d %H:%M:%S"))
+        return f"{self.movie_session.movie.title} {formatted_date} " \
+               f"(row: {self.row}, seat: {self.seat})"
```
vsmutok, line 94: better to use `()` then `\`

services/user.py
```diff
@@ -0,0 +1,46 @@
+from db.models import User
+
+
+def create_user(username: str, password: str, email: str | None = None,
+                first_name: str | None = None,
+                last_name: str | None = None) -> None:
+    new_user = User.objects.create_user(username=username, password=password)
```
vsmutok, line 7: it is better to use get_user_model() method instead of User
fix this everywhere

vsmutok (changes requested):

services/user.py
MykhailoIvchenko: Done

db/models.py
vasylhnatiuk: fix it

vasylhnatiuk (changes requested):

EdAlekseiev (approved):

PR #738, opened 2024-05-05
Solution

siaserb:


db/models.py
```diff
@@ -50,3 +59,68 @@ class MovieSession(models.Model):
...
+        return str(self.created_at)
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        MovieSession,
+        on_delete=models.CASCADE,
+        related_name="tickets",
+    )
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
```
LiudmylaKulinchenko, line 89: add related name here

db/models.py
```diff
@@ -22,6 +26,11 @@ class Movie(models.Model):
     actors = models.ManyToManyField(to=Actor, related_name="movies")
     genres = models.ManyToManyField(to=Genre, related_name="movies")
 
+    class Meta:
+        indexes = [
+            models.Index(fields=["title"], name="title_idx"),
```
LiudmylaKulinchenko, line 31: use full words here to make it more descriptive

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```
LiudmylaKulinchenko, line 36: You can use transaction decorator if you want to cover full function

LiudmylaKulinchenko: same everywhere

settings.py
```diff
@@ -17,12 +19,14 @@
...
+TIME_ZONE = "Europe/Kyiv"
 
 USE_I18N = True
 
 USE_TZ = False
 
 INSTALLED_APPS = [
     "db",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
```
LiudmylaKulinchenko, line 31: Your created apps should be the last and the default ones should be the first

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
GJ

PR #740, opened 2024-05-06
solution

nrjss67:


services/movie.py
```diff
@@ -1,14 +1,19 @@
 from django.db.models import QuerySet
+from django.db import transaction
 
 from db.models import Movie
 
 
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
EdAlekseiev, line 10: If argument can be Nony, you should show it in type hint. Fix everywhere, please
```suggestion
    title: str | None = None
```

services/movie.py
```diff
@@ -1,14 +1,19 @@
...
 ) -> QuerySet:
```
EdAlekseiev, line 11: You should specify inner type of `QuerySet`. Fix everywhere, too
```suggestion
) -> QuerySet[Movie]:
```

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
EdAlekseiev, line 36: Use `transaction.atomic` as decorator instead of context manager. It makes your code more readable.

services/order.py
```diff
@@ -0,0 +1,28 @@
+from db.models import Order, Ticket, User, MovieSession
+from django.db import transaction
+
+
+def create_order(tickets: list[dict],
+                 username: str,
+                 date: str = None) -> None:
+    with transaction.atomic():
```
EdAlekseiev, line 8: The same code as above

EdAlekseiev (changes requested):

services/movie.py
vasylhnatiuk: do it

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #741, opened 2024-05-06
Solution

EugeneKozhenevskyi:


services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
```suggestion
    title: str | None = None
```

services/movie.py
```diff
@@ -6,6 +7,7 @@
...
 ) -> QuerySet:
```
EdAlekseiev, line 11: You should specify inner type of `QuerySet`. Fix everywhere, please
```suggestion
) -> QuerySet[Movie]:
```

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```
EdAlekseiev, line 36: Use `transaction.atomic` as decorator instead of context manager. It will make your code more readable. Fix everywhere, please

EdAlekseiev (changes requested):

services/user.py
```diff
@@ -0,0 +1,47 @@
...
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+    return user
+
+
+def get_user(user_id: int) -> QuerySet[User]:
+    user = User.objects.get(pk=user_id)
```
vsmutok, line 24: it is better to use get_user_model() method instead of User
fix this everywhere

vsmutok (changes requested):

EdAlekseiev (approved):

PR #744, opened 2024-05-10
Develop

halyna-baklanova:


services/user.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def update_user(
+    user_id: int,
+    username: str = None,
+    password: str = None,
+    email: str = None,
+    first_name: str = None,
+    last_name: str = None
+) -> Union[User, None]:
```
vasylhnatiuk, line 40:
```suggestion
) -> None:

```

vasylhnatiuk: this function can return only None

vasylhnatiuk (changes requested):

services/movie_session.py
```diff
@@ -13,6 +13,13 @@ def create_movie_session(
     )
 
 
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return list(Ticket.objects.filter(
+        movie_session_id=movie_session_id).values(
+        "row", "seat")
```
Dimosphen1, line 19: In case there are no arguments on the same line with an opening parenthesis, your closing parenthesis should not have any arguments on the same line as well. Please, resolve this issue

db/models.py
```diff
@@ -50,3 +57,52 @@ class MovieSession(models.Model):
...
+                    "row number must be in available range: (1, rows): (1, {})"
+                    .format(self.movie_session.cinema_hall.rows)
+                ]
+            })
+
+        if self.seat > self.movie_session.cinema_hall.seats_in_row:
+            raise ValidationError({"seat": [
+                "seat number must be in available range: "
+                "(1, seats_in_row): (1, {})"
+                .format(self.movie_session.cinema_hall.seats_in_row)
```
Dimosphen1, line 100: It would be better to not place `.format` on a new line and leave it on the same line with string, performing line breaking after `(`

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+
+from db.models import Order, Ticket, MovieSession
+from django.db import transaction
+from django.contrib.auth import get_user_model
+
+
+def create_order(tickets: list[dict],
+                 username: str,
+                 date: datetime = None
+                 ) -> Order:
```
Dimosphen1, line 13: Place each argument on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+    with transaction.atomic():
```
Dimosphen1, line 14: In case all body of a method should be treated as an atomic transaction, it is better to use `transaction.atomic` as a decorator

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+            order.created_at = date
+            order.save()
+
+        Ticket.objects.bulk_create([Ticket(
+            movie_session=MovieSession.objects.get(
+                id=ticket["movie_session"]),
+            order=order,
+            row=ticket["row"],
+            seat=ticket["seat"]
+        ) for ticket in tickets])
```
Dimosphen1, line 28: Make correct indentation in the code

services/order.py
```diff
@@ -0,0 +1,35 @@
+from datetime import datetime
+
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket, MovieSession
+from django.db import transaction
+from django.contrib.auth import get_user_model
```
Dimosphen1, line 7: Add a blank line between different import groups and order them correctly to follow PEP-8

Dimosphen1 (changes requested):

PR #745, opened 2024-05-12
Solution

dkibalenko:


db/migrations/0001_initial.py
Dimosphen1, line 1: Do not delete the migration

db/migrations/migrations.py
Dimosphen1, line 1: Create a new migration apart from the one that is already present in the project

Dimosphen1 (changes requested):
Make sure that `flake8` tests pass

services/movie.py
```diff
@@ -28,13 +33,15 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```

services/movie_session.py
```diff
@@ -42,3 +42,10 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return [
+        {"row": ticket.row, "seat": ticket.seat}
+        for ticket in get_movie_session_by_id(movie_session_id).tickets.all()
+    ]
```
Dimosphen1, line 51: Retrieve required values without list comprehension but rather using Django ORM and `values` or `values_list` for better efficiency

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+
+from db.models import Order, Ticket, MovieSession
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
+    with transaction.atomic():
```
Dimosphen1, line 13: Same comment with atomic transaction

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+    return get_user_model().objects.get(pk=user_id)
+
+
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None
+                ) -> None:
```
Dimosphen1, line 38: Place each argument on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+                movie_session=MovieSession.objects.get(
+                    id=data["movie_session"]
+                )
+            )
+            for data in tickets
+        ]
+        Ticket.objects.bulk_create(tickets)
+
+
+def get_orders(username: str = None) -> QuerySet:
```
Dimosphen1, line 33: Indicate a `QuerySet` of what instances you return in the return type annotation

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #748, opened 2024-05-16
Develop

VasylRybalka:


services/order.py
```diff
@@ -0,0 +1,37 @@
...
+from db.models import Ticket, Order
+
+
+@transaction.atomic
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: Optional[str] = None
+) -> None:
+    with transaction.atomic():
```
EdAlekseiev, line 16: You should use `transaction.atomic` as decorator instead of context manager. It makes your code more readable

EdAlekseiev (changes requested):

PR #749, opened 2024-05-16
Solution

SerhiiDomik:


services/order.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
+    user = User.objects.get(username=username)
+
+    with transaction.atomic():
```
EdAlekseiev, line 13: You should use transaction.atomic as decorator instead of context manager. It makes your code more readable

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
```suggestion
    title: str | None = None
```

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #752, opened 2024-05-17
solution

haldaniko:


services/order.py
```diff
@@ -0,0 +1,44 @@
...
+        date: str = None
+) -> Order:
+    try:
+        user = get_user_model().objects.get(username=username)
+    except User.DoesNotExist:
+        raise ValueError(f"There is no user with username '{username}'")
+
+    with transaction.atomic():
+        order = Order.objects.create(user=user)
+
```
vsmutok, line 20: In case all body of a method should be treated as an atomic transaction, it is better to use transaction.atomic as a decorator

services/user.py
```diff
@@ -0,0 +1,60 @@
...
+
+    if last_name:
+        user.last_name = last_name
+
+    try:
+        user.full_clean()
+        user.save()
+        return user
+    except Exception as e:
+        raise ValueError(f"Error: {e}")
```
vsmutok, line 60: I think `try .. except` redundant here

vsmutok (changes requested):

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
+        movie = Movie.objects.create(
```
vsmutok, line 37: In case all body of a method should be treated as an atomic transaction, it is better to use transaction.atomic as a decorator

services/order.py
```diff
@@ -0,0 +1,44 @@
...
+from db.models import Order, User, Ticket, MovieSession
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> Order:
+    with transaction.atomic():
```
vsmutok, line 13: In case all body of a method should be treated as an atomic transaction, it is better to use transaction.atomic as a decorator

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #755, opened 2024-05-21
Solution

Viesich:


db/migrations/0003_user_order_ticket_movie_db_movie_title_5d0841_idx_and_more.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

db/models.py
```diff
@@ -50,3 +62,68 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey("Order", on_delete=models.CASCADE)
```
Dimosphen1, line 69: Add `related_name` attribute to corresponding fields

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```

services/movie_session.py
```diff
@@ -24,6 +24,11 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
     return MovieSession.objects.get(id=movie_session_id)
 
 
+def get_taken_seats(movie_session_id: int) -> list:
+    tickets = Ticket.objects.filter(movie_session_id=movie_session_id)
+    return [{"row": ticket.row, "seat": ticket.seat} for ticket in tickets]
```
Dimosphen1, line 29: Retrieve all values at once without list comprehension using Django ORM and `values` or `values_list` here

services/user.py
```diff
@@ -0,0 +1,45 @@
...
+        email: str = None,
+        first_name: str = "",
+        last_name: str = ""
+) -> None:
+    get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        email=email,
+        first_name=first_name,
+        last_name=last_name)
```
Dimosphen1, line 18: In case there are no arguments on the same line with an opening parenthesis, your closing parenthesis should not have any arguments on the same line as well. Please, resolve this issue

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+from db.models import Order, Ticket, MovieSession
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> Order:
+    user = get_user_model().objects.get(username=username)
+    with transaction.atomic():
```
Dimosphen1, line 15: Same logic with `transaction.atomic` decorator applies here as well

Dimosphen1 (changes requested):

db/migrations/0001_initial.py
```diff
@@ -1,17 +1,48 @@
-# Generated by Django 4.0.2 on 2022-06-15 12:37
+# Generated by Django 4.0.2 on 2024-05-21 14:44
 
+from django.conf import settings
+import django.contrib.auth.models
+import django.contrib.auth.validators
 from django.db import migrations, models
 import django.db.models.deletion
+import django.utils.timezone
```
pavlopro, line 10: DO not modify this migration. Add new one.

It is a good practice to include one migration per app per PR in Django projects since it helps to avoid migrations folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

db/models.py
```diff
@@ -50,3 +62,76 @@ class MovieSession(models.Model):
...
+
+    def __str__(self) -> str:
+        return (
+            f"Matrix {self.movie_session.show_time.date()} "
+            f"{self.movie_session.show_time.time()} "
+            f"(row: {self.row}, seat: {self.seat})"
+        )
+
+    def clean(self) -> None:
+        if self.row > self.movie_session.cinema_hall.rows:
```
pavlopro, line 89: Also include into check that row is not < 1

db/models.py
```diff
@@ -50,3 +62,76 @@ class MovieSession(models.Model):
...
+            raise ValidationError(
+                {
+                    "row": [
+                        f"row number must be in available range: (1, rows): "
+                        f"(1, {self.movie_session.cinema_hall.rows})"
+                    ]
+                }
+            )
+        if self.seat > self.movie_session.cinema_hall.seats_in_row:
```
pavlopro, line 98: Also include into check that seat is not < 1

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+            row=ticket_data["row"],
+            seat=ticket_data["seat"],
+        )
+    if date:
+        order.created_at = date
+        order.save()
+    return order
+
+
+def get_orders(username: str = None) -> Order:
```
pavlopro, line 33: QuerySet[Order]

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+    orders = Order.objects.all()
+    if username:
```
pavlopro, line 35:
```
if username:
```

vsmutok (approved):
Great!

PR #763, opened 2024-06-09
Solution

sind14:


services/order.py
```diff
@@ -0,0 +1,33 @@
...
+            movie_session_id = ticket["movie_session"]
+            movie_session = MovieSession.objects.get(id=movie_session_id)
+            Ticket.objects.create(
+                movie_session=movie_session,
+                order=order,
+                row=row,
+                seat=seat
+            )
+    order.save()
+    return order
```
vsmutok, line 26: use transaction.atomic decorator. The decorator is a more concise and cleaner way to ensure that the function executes within a transaction.

vsmutok (changes requested):

EdAlekseiev (approved):

PR #767, opened 2024-06-13
Solution

CrazyHilly:


services/movie.py
```diff
@@ -6,16 +7,17 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
EdAlekseiev, line 10: If argument can be `None`, you should show it in type hint. Fix everywhere, please
```suggestion
    title: str | None = None,
```

services/movie.py
```diff
@@ -28,13 +30,15 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
 
-    return movie
+    with transaction.atomic():
```
EdAlekseiev, line 34: Use transaction.atomic as decorator insteead of context manager. It makes your code more readable

services/order.py
```diff
@@ -0,0 +1,33 @@
+from django.contrib.auth import get_user_model
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket, MovieSession
+
+
+def create_order(tickets: list[dict], username: str, date: str = None) -> None:
+    with transaction.atomic():
```
EdAlekseiev, line 9: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #770, opened 2024-06-16
Solution

SuskyiVolodymyr:


services/movie.py
```diff
@@ -15,6 +17,9 @@ def get_movies(
     if actors_ids:
         queryset = queryset.filter(actors__id__in=actors_ids)
 
+    if title:
+        queryset = queryset.filter(title__icontains=title)
+
     return queryset
```
Oleg-s-s, line 23: There is also more advanced and db friendly approach:
```suggestion
    query = Q()
    if ...:
        query.add(Q(), Q.AND)
    queryset = Movie.objects.filter(query)
```

Oleg-s-s (changes requested):

DevSpaciX (approved):

PR #771, opened 2024-06-17
Implement solution

anton-bilous:


anton-bilous:
Note that I've used `default=timezone.now` instead of `auto_now_add=True`, because `auto_now_add` can not be overriden and the tests try to override it.
From django documentation:

> Note that the current date is always used; it’s not just a default value that you can override. So even if you set a value for this field when creating the object, it will be ignored.

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
...
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return [
+        {"row": ticket.row, "seat": ticket.seat}
+        for ticket in MovieSession.objects.get(
+            id=movie_session_id,
+        ).tickets.all()
+    ]
```
LiudmylaKulinchenko, line 53: You can simplify it by using `filter` instead of `get` + using `values`

services/user.py
```diff
@@ -0,0 +1,62 @@
...
+        kwargs["first_name"] = first_name
+
+    if last_name:
+        kwargs["last_name"] = last_name
+
+    return get_user_model().objects.create_user(
+        username,
+        password=password,
+        **kwargs,
+    )
```
LiudmylaKulinchenko, line 28: Looks like in this case there's no need to create kwargs dictionary, what do you think?

LiudmylaKulinchenko (changes requested):
Note that I've used default=timezone.now instead of auto_now_add=True, because auto_now_add can not be overriden and the tests try to override it` - that's weird, it should work with auto_now_add. Let's try to use it again! There's probably an issue if it doesn't work

oliasymak24 (approved):

PR #778, opened 2024-06-26
2 new models and some new services

Timur5050:


Timur5050:
why commits are not from my account? because i use my friends laptop for some time because of problems with electricity

db/migrations/0004_alter_ticket_order.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

db/models.py
```diff
@@ -50,3 +58,62 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE
```
Dimosphen1, line 67: Add a `related_name` here as well

services/order.py
```diff
@@ -0,0 +1,45 @@
...
+
+from db.models import Order, Ticket, User
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
+    with (transaction.atomic()):
```
Dimosphen1, line 14: In case all body of a method should be treated as an atomic transaction, it is better to use `transaction.atomic` as a decorator. Apply this change to the `create_movie` as well

services/order.py
```diff
@@ -0,0 +1,45 @@
...
+        ]
+        for ticket in tickets_objects:
+            ticket.clean()
+
+        Ticket.objects.bulk_create(tickets_objects)
+
+
+def get_orders(
+        username: str = None
+) -> QuerySet:
```
Dimosphen1, line 40: Indicate a `QuerySet` of what instances you return in the return type annotation

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> User:
+    new_user = User.objects.create_user(
```
Dimosphen1, line 11: Use `get_user_model` instead of `User` here and in the other places as well as mentioned in the checklist

Dimosphen1 (changes requested):

pavlopro (changes requested):

db/migrations/0001_initial.py
Dimosphen1, line 1: Do not delete or redefine the existing migration. You should create a new one upon the migration that is already present in the project to preserve a consistent state in the DB

db/models.py
```diff
@@ -50,3 +58,63 @@ class MovieSession(models.Model):
...
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique rows, seats and movie_sessions"
+            )
+        ]
+
+    def clean(self) -> bool:
```
Dimosphen1, line 100: Indicate in the return type annotation that you return `None`, not `bool`

services/movie_session.py
```diff
@@ -42,3 +42,9 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
+    session = MovieSession.objects.get(pk=movie_session_id)
+    all_seats = session.tickets.all()
+    return [{"row": place.row, "seat": place.seat} for place in all_seats]
```
Dimosphen1, line 50: You could retrieve these values using Django ORM and `values` without iterating over objects in the list comprehension

Dimosphen1 (changes requested):

Timur5050:
i dont understand about problems with that migrations, i dont see any problems

EdAlekseiev (approved):

PR #781, opened 2024-06-28
Solution

VladimirDolhyi:


VladimirDolhyi:
> Add one migrations per PR

Ok.

PR #782, opened 2024-06-29
Solution

Unlie9:


services/user.py
```diff
@@ -0,0 +1,43 @@
...
+        user.last_name = last_name
+    if email:
+        user.email = email
+
+    user.save()
+    return user
+
+
+def get_user(user_id: int) -> int:
+    return User.objects.get(id=user_id)
```
DevSpaciX, line 22: Try to use `get_user_model()`
And why return type here is `int` ?))

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+        if date:
+            order.created_at = date
+
+        ticket_order = [
+            Ticket(
+                row=ticket["row"],
+                seat=ticket["seat"],
+                movie_session=MovieSession.objects.get(
+                    id=ticket["movie_session"]
+                ),
```
DevSpaciX, line 23: Use movie_session_id argument and pass the id directly.

services/order.py
```diff
@@ -0,0 +1,36 @@
+from typing import List, Dict, Any
+
+from django.db import transaction
+from django.contrib.auth import get_user_model
+
+from db.models import Order, Ticket, User, MovieSession
+
+
+def create_order(tickets: Ticket, username: User, date: str = None) -> str:
+    with transaction.atomic():
```
DevSpaciX, line 10: Fix return type hints everywhere : )

db/models.py
```diff
@@ -50,3 +53,53 @@ class MovieSession(models.Model):
...
+    class Meta:
+        ordering = ("-created_at",)
+
+    def __str__(self) -> str:
+        return f"{self.created_at.strftime('%Y-%m-%d %H:%M:%S')}"
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey("MovieSession", on_delete=models.CASCADE)
+    order = models.ForeignKey("Order", on_delete=models.CASCADE)
```
DevSpaciX, line 71: Add related name param : )

db/models.py
```diff
@@ -50,3 +53,53 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey("User", on_delete=models.CASCADE)
```
DevSpaciX, line 60: you should refer to AUTH_USER_MODEL

DevSpaciX (changes requested):

services/user.py
```diff
@@ -0,0 +1,44 @@
...
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(user_id: int, username: str = None,
+                password: str = None, first_name: str = None,
+                last_name: str = None, email: str = None) -> None:
+    try:
+        user = User.objects.get(id=user_id)
+    except User.DoesNotExist:
+        raise ValueError(f"User with id {user_id} does not exist")
```
DevSpaciX, line 32: Could you use here get_user() func ?

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+            )
+            for ticket in tickets
+        ]
+        order.save()
+        Ticket.objects.bulk_create(ticket_order)
+        return order
+
+
+def get_orders(username: User = None) -> list[Order]:
+    if username:
```
DevSpaciX, line 38: Remember:`.all()`, `.filter()` returns QuerySet`.get()`, `.delete()`, `.create()` returns model class, `-> User` for. example

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #783, opened 2024-06-30
Solution

BohdanKuzik:


services/order.py
```diff
@@ -0,0 +1,38 @@
...
+            ticket.full_clean()
+            ticket.save()
+
+        return order
+
+
+def get_orders(username: User = None) -> None:
+    if username:
+        user = get_user_model().objects.get(username=username)
+        orders = Order.objects.filter(user=user)
```
DevSpaciX, line 34: Could you use here this filter ?
`user__username=username`

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+    else:
+        orders = Order.objects.all()
+
+    return orders
```
DevSpaciX, line 38: Also try to do this without `else` stmt

services/movie_session.py
```diff
@@ -41,4 +41,15 @@ def update_movie_session(
...
+
+def get_taken_seats(movie_session_id: int) -> list:
+    movie_session = get_movie_session_by_id(movie_session_id)
+    if not movie_session:
+        return []
+    tickets = movie_session.tickets.all()
+    return [
+        {"row": ticket.row, "seat": ticket.seat}
+        for ticket in tickets
+    ]
```
DevSpaciX, line 55: `if not movie_session` block can be removed ?

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #784, opened 2024-06-30
Solution

Yuruch:


db/models.py
```diff
@@ -50,3 +60,67 @@ class MovieSession(models.Model):
...
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket_row_seat_movie_session"
+            )
+        ]
```
DevSpaciX, line 122: class Meta should be placed before methods

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #793, opened 2024-07-01
Solution

VShveda:


services/order.py
```diff
@@ -0,0 +1,43 @@
...
+        return order
+
+
+def get_orders(username: User = None) -> QuerySet:
+    if username:
+        user = get_user_model().objects.get(username=username)
+        orders = Order.objects.filter(user=user)
+    else:
+        orders = Order.objects.all()
+    return orders
```
DevSpaciX, line 43: Could you filter order by this condition ?
`user__username=username`
Also try to remove `else` stmt

services/order.py
```diff
@@ -0,0 +1,43 @@
...
+
+        for ticket_data in tickets:
+            movie_session = MovieSession.objects.get(
+                pk=ticket_data["movie_session"]
+            )
+            ticket = Ticket(
+                row=ticket_data["row"],
+                seat=ticket_data["seat"],
+                movie_session=movie_session,
+                order=order
```
DevSpaciX, line 29: you could set movie session just by id which is stored in `ticket_data["movie_session"]`

db/models.py
```diff
@@ -50,3 +57,69 @@ class MovieSession(models.Model):
...
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=("row", "seat", "movie_session"),
+                name="unique_ticket"
+            )
+        ]
```
DevSpaciX, line 121: class Meta should be placed before methods

db/models.py
```diff
@@ -50,3 +57,69 @@ class MovieSession(models.Model):
...
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        "User",
+        on_delete=models.CASCADE
+    )
```
DevSpaciX, line 67: you should refer to AUTH_USER_MODEL

DevSpaciX (changes requested):

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #795, opened 2024-07-01
sol1

IvanVahanov:


services/user.py
```diff
@@ -0,0 +1,39 @@
...
+        password=password,
+        first_name=first_name,
+        last_name=last_name,
+        email=email
+    )
+    return user
+
+
+def get_user(user_id: int) -> int:
+    return User.objects.get(id=user_id)
```
DevSpaciX, line 21: Use better `get_user_model()` here

services/user.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def update_user(user_id: int, username: str = None,
+                password: str = None, email: str = None,
+                first_name: str = None, last_name: str = None) -> QuerySet:
+    user = User.objects.get(id=user_id)
```
DevSpaciX, line 27: Could you reuse your `get_user `func here ?

db/models.py
```diff
@@ -50,3 +64,54 @@ class MovieSession(models.Model):
...
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=["movie_session", "row", "seat"],
+                                    name="unique_ticket")
+        ]
```
DevSpaciX, line 117: class Meta should be placed before methods

db/models.py
```diff
@@ -50,3 +64,54 @@ class MovieSession(models.Model):
...
+    def save(self, *args, **kwargs) -> None:
+        if not self.created_at:
+            self.created_at = timezone.now()
+        super().save(*args, **kwargs)
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+    class Meta:
+        ordering = ["-created_at"]
```
DevSpaciX, line 83: class Meta should be placed before methods

DevSpaciX (changes requested):

PR #796, opened 2024-07-01
Solution

YevaYe:


services/user.py
```diff
@@ -0,0 +1,56 @@
...
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None,
+) -> User:
+    user = get_user_model().objects.get(pk=user_id)
+
```
DevSpaciX, line 43: Could you reuse here `get_user()` func ?

db/models.py
```diff
@@ -50,3 +53,65 @@ class MovieSession(models.Model):
...
+    def save(self, *args, **kwargs) -> callable:
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_row_seat_movie_session"),
+        ]
```
DevSpaciX, line 117: class Meta should be placed before methods

db/models.py
```diff
@@ -50,3 +53,65 @@ class MovieSession(models.Model):
...
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        to=User,
```
DevSpaciX, line 65: you should refer to AUTH_USER_MODEL

DevSpaciX (changes requested):

services/user.py
YevaYe: This method we used the video

settings.py
```diff
@@ -25,4 +25,8 @@
 
 INSTALLED_APPS = [
     "db",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
```
Arsen-hrynevych, line 29: Could you clarify the purpose? I'm not sure I understand.

services/user.py
```diff
@@ -0,0 +1,56 @@
+from django.contrib.auth import get_user_model
+
+from db.models import User
```
Arsen-hrynevych, line 3: The import from db.models import User is redundant and potentially problematic. You're using Django's get_user_model() to fetch the user model, so importing User from your db.models isn't necessary and could cause confusion if User doesn't match the one returned by get_user_model()

services/user.py
Arsen-hrynevych: user = User.objects.get(pk=user_id)

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+
```
Arsen-hrynevych, line 4: User = get_user_model()

Arsen-hrynevych (changes requested):
Mentors and videos don't guarantee that what you see is the best solution. Don't treat them as the definitive or 100% correct answers.

settings.py
YevaYe: I had it like one of tasks in README.md

services/user.py
YevaYe: I've done it but I am not sure if I understood you correctly

Arsen-hrynevych (approved):
Thanks

PR #797, opened 2024-07-01
Solution

Andrii1000:


services/order.py
```diff
@@ -0,0 +1,34 @@
...
+    user = get_user_model().objects.get(username=username)
+    order = Order.objects.create(user=user)
+    if date:
+        order.created_at = date
+    order.save()
+
+    for ticket in tickets:
+        movie_session = MovieSession.objects.get(id=ticket["movie_session"])
+        Ticket.objects.create(
+            movie_session=movie_session,
```
DevSpaciX, line 23: it is better to set movie_session by id 
Use `movie_session_id` argument and pass the id directly.

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #798, opened 2024-07-01
solution

AntonKorinchuk:


services/user.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> User:
+    user = get_user_model().objects.get(id=user_id)
+    if username:
```
DevSpaciX, line 34: Just reuse your `get_user` func )

DevSpaciX (changes requested):

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+    if date:
+        order.created_at = date
+    Ticket.objects.bulk_create(
+        [
+            Ticket(
+                row=ticket["row"],
+                seat=ticket["seat"],
+                movie_session=MovieSession.objects.get(
+                    id=ticket["movie_session"]
+                ),
```
DevSpaciX, line 27: Sorry did not see that in first time . 
Use movie_session_id argument and pass the id directly.

db/models.py
```diff
@@ -50,3 +57,68 @@ class MovieSession(models.Model):
...
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_row_seat_movie_session"),
+        ]
```
DevSpaciX, line 120: class Meta should be placed before methods

DevSpaciX (changes requested):

DevSpaciX (changes requested):
Use `movie_session_id` argument and pass the id directly.

DevSpaciX (approved):

PR #799, opened 2024-07-01
Develop

b3v3kt0r:


services/order.py
```diff
@@ -0,0 +1,26 @@
...
+            Ticket.objects.create(order_id=order.id,
+                                  row=ticket["row"],
+                                  seat=ticket["seat"],
+                                  movie_session=session)
+
+
+def get_orders(username: str = None) -> Order | QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
+    return Order.objects.all()
```
DevSpaciX, line 26: Remember:`.all()`, `.filter()` returns QuerySet`.get()`, `.delete()`, `.create()` returns model class, `-> User` for. example

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #800, opened 2024-07-01
Solution

lilarin:


db/models.py
```diff
@@ -22,10 +24,79 @@ class Movie(models.Model):
...
         return self.title
 
 
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey("User", on_delete=models.CASCADE)
```
DevSpaciX, line 42: you should refer to AUTH_USER_MODEL

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+        order = Order.objects.create(user=user)
+
+        if date:
+            order.created_at = date
+            order.save()
+
+        for ticket_data in tickets:
+            movie_session = MovieSession.objects.get(
+                id=ticket_data["movie_session"]
+            )
```
DevSpaciX, line 23: Use movie_session_id argument and pass the id directly.

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #801, opened 2024-07-02
'Solution'

Irina17191:


services/user.py
```diff
@@ -0,0 +1,51 @@
...
+
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> None:
+
+    user = get_user_model().objects.get(pk=user_id)
+
```
DevSpaciX, line 39: Could you just reuse get_user() func ?

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+            )
+
+        order.save()
+        return order
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        user = get_user_model().objects.get(username=username)
+        return Order.objects.filter(user=user)
```
DevSpaciX, line 40: Could you use here this filter ?
`user__username=username`

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+        for ticket in tickets:
+            movie_session_id = ticket.get("movie_session")
+            row = ticket.get("row")
+            seat = ticket.get("seat")
+
+            movie_session = MovieSession.objects.get(pk=movie_session_id)
+
+            Ticket.objects.create(
+                order=order,
+                movie_session=movie_session,
```
DevSpaciX, line 28: Use movie_session_id argument and pass the id directly.

services/movie_session.py
```diff
@@ -42,3 +41,17 @@ def update_movie_session(
...
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+
+    taken_seats = []
+    all_taken_tickets = Ticket.objects.filter(movie_session=movie_session_id)
+
+    for ticket in all_taken_tickets:
+        taken_seats.append({
+            "row": ticket.row,
+            "seat": ticket.seat,
+        })
```
DevSpaciX, line 55: Could you use here dict comprehension ?

db/models.py
```diff
@@ -50,3 +54,61 @@ class MovieSession(models.Model):
...
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(fields=["row", "seat", "movie_session"],
+                             name="unique_movie_session"),
+        ]
```
DevSpaciX, line 110: class Meta should be placed before methods

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #805, opened 2024-07-05
Solution

VitaliiDats1:


DevSpaciX (changes requested):
Push the migrations and fix the tests : ) They are failed now

services/user.py
```diff
@@ -0,0 +1,50 @@
...
+    return get_user_model().objects.get(pk=user_id)
+
+
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> None:
+    user = User.objects.get(id=user_id)
```
DevSpaciX, line 38: Reuse here `get_user()` func

services/user.py
```diff
@@ -0,0 +1,50 @@
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
+
```
DevSpaciX, line 4: Remove this )

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+                seat=ticket_data["seat"]
+            )
+
+    return order
+
+
+def get_orders(username: str = None) -> list[dict]:
+    if username:
+        user = User.objects.get(username=username)
+        return Order.objects.filter(user=user)
```
DevSpaciX, line 33: Could you use here this filter ?
`user__username=username`

DevSpaciX (changes requested):

PR #819, opened 2024-07-18
solution

AndreiVed:


services/order.py
```diff
@@ -0,0 +1,33 @@
...
+from db.models import Order, Ticket
+
+
+def create_order(
+    tickets: list[dict],
+    username: str,
+    date: str = None
+) -> None:
+
+    with transaction.atomic():
```
vasylhnatiuk, line 13: use @transaction.atomic instead of with

vasylhnatiuk (changes requested):

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+    for ticket in tickets:
+        Ticket.objects.create(
+            movie_session_id=ticket["movie_session"],
+            row=ticket["row"],
+            seat=ticket["seat"],
+            order=order
+        )
+
+
+def get_orders(username: str = None) -> Order:
```
vasylhnatiuk, line 29:
```suggestion
def get_orders(username: str = None) -> QuerySet:
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #820, opened 2024-07-19
implement py-tickets-and-orders

olenazaritska:


services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return [
```
pavlopro, line 48: Import `Ticket `and filter by movie_session_id. Use` .values() `method

pavlopro (changes requested):

PR #823, opened 2024-07-21
Solution 21/07/24

Viva1-ua:


vsmutok (changes requested):
Tests Failed!

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+            order.save()
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"], order=order,
+                row=ticket["row"], seat=ticket["seat"]
+            )
+
+        order.save()
+    return order
```
vsmutok, line 27: why do you need `with` here if you have a `@transaction.atomic`

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
+        movie = Movie.objects.create(
+            title=movie_title,
+            description=movie_description,
+        )
+        if genres_ids:
+            movie.genres.set(genres_ids)
+        if actors_ids:
+            movie.actors.set(actors_ids)
```
vsmutok, line 44: use `@transaction.atomic` instead of with 
this will make your code more readable and clean

vsmutok (changes requested):

EdAlekseiev (approved):

PR #834, opened 2024-07-30
Develop

DaniaK123:


services/movie.py
```diff
@@ -28,13 +31,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
EdAlekseiev, line 34: Use transaction.atomic as decorator instead of context manager. It will make your code more readable

EdAlekseiev (changes requested):

PR #835, opened 2024-08-04
Solution

raychw:


services/order.py
```diff
@@ -0,0 +1,42 @@
+from django.contrib.auth import get_user_model
+
```
Dimosphen1, line 2: The blank line should be after `from django.db import transaction` import since it relates to the Django framework

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #839, opened 2024-08-06
Implemented solution

dimak20:


db/models.py
```diff
@@ -1,5 +1,11 @@
+import datetime
+
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
 from django.db import models
 
+import settings
```
`from django.conf import settings` because it defines correct settings(as there may be a folder named settings)

db/models.py
```diff
@@ -17,14 +23,19 @@ def __str__(self) -> str:
 
 
 class Movie(models.Model):
-    title = models.CharField(max_length=255)
+    title = models.CharField(max_length=255, db_index=True)
```
MNDonut, line 26: either db_index = True or indexes should be used. indexes are often used for complex indexes

db/models.py
```diff
@@ -50,3 +61,69 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(default=datetime.datetime.now)
```
MNDonut, line 67: use auto_now_add instead

db/models.py
```diff
@@ -50,3 +61,69 @@ class MovieSession(models.Model):
...
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.CASCADE,
```
MNDonut, line 69: we don't add multiple parameters on the same line

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
MNDonut, line 10: str or None

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket, MovieSession
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
```
MNDonut, line 11: str or None

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+                row=ticket["row"],
+                seat=ticket["seat"],
+                order=order
+            )
+            for ticket in tickets
+        ]
+        Ticket.objects.bulk_create(raw_tickets)
+
+
+def get_orders(username: str = None) -> QuerySet:
```
MNDonut, line 38: str or None

services/user.py
```diff
@@ -0,0 +1,45 @@
+from django.contrib.auth import get_user_model
+
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
```
MNDonut, line 9: str or None

services/user.py
```diff
@@ -0,0 +1,45 @@
...
+    return get_user_model().objects.get(id=user_id)
+
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
```
MNDonut, line 32: str or None

MNDonut (changes requested):

db/models.py
dimak20: I tried but It did not work due to mocking datetime. And It would be too strange to override save methods just to pass all the tests

MNDonut (approved):

PR #849, opened 2024-08-21
"Solution"

OleksiukStepan:


db/models.py
```diff
@@ -1,4 +1,9 @@
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
 from django.db import models
+from django.db.models import UniqueConstraint
+
+import settings
```
MNDonut, line 6: it's better to use 
`from django.conf import settings` because you may have a few settings files or a folder named "settings". This import searches for real settings

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
MNDonut, line 10: list/str or None type. Fix annotation everywhere

services/movie.py
```diff
@@ -15,13 +17,17 @@ def get_movies(
     if actors_ids:
         queryset = queryset.filter(actors__id__in=actors_ids)
 
+    if title:
+        queryset = queryset.filter(title__contains=title)
```
MNDonut, line 21: I recommend to use "icontains" because when you look for the movie "The Maze Runner" and enter "maze" it should find the movie as well(for user's comfortability)
So:
contains - case-sensitive (when Maze and maze are different words)
**i**contains - case-**i**nsensitive (when both Maze and maze are the same)

MNDonut (changes requested):

oliasymak24 (approved):

PR #850, opened 2024-08-22
Solution

dartomOOv:


services/movie_session.py
```diff
@@ -42,3 +42,9 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
+    session = MovieSession.objects.get(id=movie_session_id)
+    taken_seats = session.ticket_set.values("row", "seat")
```
LiudmylaKulinchenko, line 49: You can make it in one request: filter tickets by movie_session_id

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+from db.models import Order, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None,
+) -> None:
+
+    the_user = get_user_model().objects.get(username=username)
```
LiudmylaKulinchenko, line 13: The prefix `the_` is not necessary. Simply using `user` would be more conventional and clear

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+    with transaction.atomic():
+        the_order = Order.objects.create(
```
LiudmylaKulinchenko, line 15: same for order

services/order.py
LiudmylaKulinchenko: if you want to make it more clear, you can use `order_date` instead of `date`

LiudmylaKulinchenko (changes requested):

oliasymak24 (approved):

PR #855, opened 2024-08-26
Solution

DSahalatyi:


db/models.py
```diff
@@ -1,5 +1,11 @@
+from datetime import datetime
+
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
 from django.db import models
 
+import settings
```
`from django.conf import settings` because there may be a few settings or a folder named "settings"

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
MNDonut, line 10: str or None. fix annotation everywhere

services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    movie_session = MovieSession.objects.get(pk=movie_session_id)
+    return [
+        {"row": ticket.row, "seat": ticket.seat}
+        for ticket in movie_session.tickets.all()
```
MNDonut, line 51: you can use .values() method instead of creating dicts manually

MNDonut (changes requested):

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = "",
+        last_name: str = "",
+) -> User:
+    return get_user_model().objects.create_user(
```
oliasymak24, line 13: You don’t need to `return` anything in create/update/delete methods, check everywhere

oliasymak24 (changes requested):

MNDonut (approved):

PR #856, opened 2024-08-28
'Solution'

Vitalii-Haponiuk:


services/order.py
```diff
@@ -0,0 +1,33 @@
+from db.models import Order, Ticket
```
LLkaia, line 1: place this import after all others and blank line

LLkaia: You can read about proper order of import sorting

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+        for ticket in tickets:
+            Ticket.objects.create(
+                order=order,
+                movie_session_id=ticket["movie_session"],
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
```
LLkaia, line 30:
```suggestion
def get_orders(username: str | None = None) -> QuerySet[Order]:

```

services/user.py
```diff
@@ -0,0 +1,49 @@
+from db.models import User
+from django.contrib.auth import get_user_model
```
LLkaia, line 2: Fox imports here as well

LLkaia (changes requested):

db/migrations/0001_initial.py
```diff
@@ -1,58 +1,114 @@
-# Generated by Django 4.0.2 on 2022-06-15 12:37
```
pavlopro, line 1: You shouldn't modify existing migration file
Create a new one instead

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+
+def create_user(
+    username: str,
+    password: str,
+    email: str | None = None,
+    first_name: str = "",
+    last_name: str = ""
+) -> User:
+
+    user = User.objects.create_user(
```
pavlopro, line 13: No need to have this variable
just return it instead

pavlopro (changes requested):

PR #857, opened 2024-08-28
implement tickets and orders

vladislav-tsybuliak1:


services/order.py
```diff
@@ -0,0 +1,39 @@
...
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+        return order
```
oliasymak24, line 30: You don’t need to `return` anything in create/update/delete methods, check everywhere

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+        user.username = username
+    if password:
+        user.set_password(password)
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
```
oliasymak24, line 49: Try to use update function here

settings.py
oliasymak24, line 1: Don't need to commit the file

oliasymak24 (changes requested):

services/user.py
vladislav-tsybuliak1: In Django [documentation ](https://docs.djangoproject.com/en/5.1/topics/db/queries/) it is written:
_Sometimes you want to set a field to a particular value for all the objects in a QuerySet. You can do this with the update() method._
And these:
_To save changes to an object that’s already in the database, use save()._

I think in this case, because we update only 1 object, it is better to use save().

Also, with the update function I see two possible ways how to implement it here and problems that arise after:
First, I can collect all changes in a dict and make one update in the end. I will have more code here, And I will need to create an additional dict and then itterate over it. It won't be as efficient as the code above.
Second, I can update everytime I check an argument - this will create many queries to the database If I want to change many fields.. (1 query per field). I thinks it's also innefficient.

If I'm wrong, looking forward to read from you.)

db/models.py
```diff
@@ -1,4 +1,9 @@
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
 from django.db import models
+from django.db.models import UniqueConstraint
+
+import settings
```
MNDonut, line 6: it's better to use
`from django.conf import settings` because it defines main settings for sure.
It's possible that you have multiple settings files or a folder named "settings", thus python will incorrectly import settings

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
MNDonut, line 10: str or None type. fix annotation everywhere

services/movie_session.py
```diff
@@ -42,3 +42,12 @@ def update_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    movie_session = MovieSession.objects.get(id=movie_session_id)
+    return [
+        {"row": ticket.row, "seat": ticket.seat}
+        for ticket
+        in movie_session.tickets.all()
```
MNDonut, line 52: it's better to use .values() method so that you avoid manually creating dicts

MNDonut (changes requested):

services/movie.py
vladislav-tsybuliak1: Okay, but this code was written before and in the task description I didn't find anything about editing it. But I have noticed this annotation mistakes.

services/cinema_hall.py
```diff
@@ -8,8 +8,12 @@ def get_cinema_halls() -> QuerySet:
 
 
 def create_cinema_hall(
-    hall_name: str, hall_rows: int, hall_seats_in_row: int
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
 ) -> CinemaHall:
     return CinemaHall.objects.create(
```
oliasymak24, line 15: You don’t need to `return` anything in create/update/delete methods, check everywhere

services/user.py
oliasymak24: I just want you try to use update method, because this also an way to learn it better)
Collect all changes

oliasymak24 (changes requested):

services/cinema_hall.py
vladislav-tsybuliak1: okay, but this code was written not by me)
And in the task description it is nowhere said that I need to change the existing code.
Maybe, it is better to mention it in the task.

services/movie.py
```diff
@@ -1,11 +1,13 @@
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+        genres_ids: list[int] | None = None,
```
Oleksl888, line 8: fix indentation here

services/movie.py
```diff
@@ -23,18 +28,17 @@ def get_movie_by_id(movie_id: int) -> Movie:
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+        movie_title: str,
```
Oleksl888, line 31: this is also overindented

services/movie_session.py
```diff
@@ -1,19 +1,21 @@
...
+from django.db.models import QuerySet, F
 
 from db.models import MovieSession
 
 
 def create_movie_session(
-    movie_show_time: str, movie_id: int, cinema_hall_id: int
-) -> MovieSession:
-    return MovieSession.objects.create(
+        movie_show_time: str,
```
Oleksl888, line 7: indentation

services/movie_session.py
```diff
@@ -25,10 +27,10 @@ def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
 
 
 def update_movie_session(
-    session_id: int,
-    show_time: str = None,
-    movie_id: int = None,
-    cinema_hall_id: int = None,
+        session_id: int,
```
Oleksl888, line 30: overindented

services/movie_session.py
```diff
@@ -42,3 +44,13 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return list(
```
Oleksl888, line 50: or you can do this `list(get_movie_session_by_id(movie_session_id).tickets.values("row", "seat"))`

services/user.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def get_user(user_id: int) -> get_user_model():
+    return get_user_model().objects.get(id=user_id)
```
Oleksl888, line 25: use pk here

Oleksl888 (changes requested):

MNDonut (approved):

PR #862, opened 2024-09-01
first_try

yashkunn:


Oleksl888 (changes requested):

services/user.py
```diff
@@ -0,0 +1,46 @@
...
+        username=username,
+        password=password,
+        email=email,
+        first_name=first_name,
+        last_name=last_name
+    )
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
```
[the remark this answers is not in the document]
yashkunn: чому тілки в цьому місці краще використовувати "pk"? В коді є багато моментів, де я звертаюсь до "id"

services/user.py
```diff
@@ -0,0 +1,46 @@
...
+        username=username,
+        password=password,
+        email=email,
+        first_name=first_name,
+        last_name=last_name
+    )
+
+
+def get_user(user_id: int) -> User:
+    return get_user_model().objects.get(pk=user_id)
```
MNDonut, line 23: @yashkunn Regarding your question why pk only here
it's always preferable to use pk over id because in the future you may change the primary key (for example name that is unique instead of id). In this case everything will break cuz you hardcoded id field instead of dynamic pk. Oleksandr mentioned it once so that you know about this issue and can fix it everywhere ;)

settings.py
```diff
@@ -25,4 +27,6 @@
 
 INSTALLED_APPS = [
     "db",
+    "django.contrib.contenttypes",
+    "django.contrib.auth",
```
MNDonut, line 31: actually default apps should be at the top of this list and then your apps and then third-party apps

MNDonut (changes requested):

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str | None = None
+) -> Order:
+    user = get_user_model().objects.get(username=username)
+    order = Order.objects.create(user=user)
+    if date:
+        order.created_at = date
+        order.save(update_fields=["created_at"])
```
Oleksl888, line 18: you can just call save() without extra params here

services/user.py
```diff
@@ -0,0 +1,46 @@
+from django.contrib.auth import get_user_model
+
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str | None = None,
+        first_name: str = "",
```
Oleksl888, line 10: you still have to annotate that this parameter is optional here

services/user.py
```diff
@@ -0,0 +1,46 @@
...
+        last_name: str = "",
+) -> None:
+    get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        email=email,
+        first_name=first_name,
```
Oleksl888, line 17: This is possible approach, but to be safe it is better to write optional parameters to DB only when they are actually being passed. For example here you fill out first name and last name with empty strings. But what if some other part of our application expects those fields to be NULL since we know we didn't send any names there. It will lead to your DB containing some garbage values such as empty strings which may in turn affect other queries where we filter by NULL values. So in this scenario it will be better to check if param is present before you write to db

Oleksl888 (changes requested):

db/models.py
```diff
@@ -67,7 +93,7 @@ class Meta:
 
     def clean(self) -> None:
         cinema_hall = self.movie_session.cinema_hall
-        if self.row > cinema_hall.rows:
+        if self.row < 1 or self.row > cinema_hall.rows:
```
MNDonut, line 96: do you validate < 1 for seats?

services/user.py
```diff
@@ -7,16 +7,21 @@ def create_user(
...
-        password=password,
-        email=email,
-        first_name=first_name,
-        last_name=last_name
-    )
+    user_data = {
+        "username": username,
+        "password": password,
+    }
+    if email is not None:
```
MNDonut, line 17: or just "if email"

MNDonut (changes requested):

MNDonut (approved):

PR #863, opened 2024-09-01
implement py-tickets-and-orders

ValeriiaKyr:


db/models.py
```diff
@@ -50,3 +62,61 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE
```
Oleksl888, line 71: related name missing

db/models.py
```diff
@@ -50,3 +62,61 @@ class MovieSession(models.Model):
...
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        date = self.created_at.strftime("%Y-%m-%d %H:%M:%S")
+        return date
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
```
Oleksl888, line 83: related names missing

db/models.py
```diff
@@ -50,3 +62,61 @@ class MovieSession(models.Model):
...
+                {"row": f"row number must be in "
+                    f"available range: (1, rows): (1, "
+                    f"{self.movie_session.cinema_hall.rows})"}
+            )
+
+    def save(self, *args, **kwargs) -> Any:
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
```
Oleksl888, line 112: as a convention in django Meta class is defined right after fields definition

services/movie_session.py
```diff
@@ -42,3 +44,11 @@ def update_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    row_seats = Ticket.objects.filter(
+        movie_session_id=movie_session_id
+    ).values("row", "seat")
+
+    return list(row_seats)
```
Oleksl888, line 54: no need to store row_seats in a variable - just return whole expression in list()

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+from db.models import Ticket, Order
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str | None = None
+) -> Order:
+
+    user = get_user_model().objects.get(username=username)
```
Oleksl888, line 15: shouldn't this be as well under `atomic` decorator?

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+
+from django.contrib.auth import get_user_model
+
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: Any,
+        email: str = None,
```
Oleksl888, line 11: add annotation that these params are optional

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
```
Oleksl888, line 33: use get_user_model() here. Also use pk instead of id

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def update_user(
+        user_id: int,
+        username: str | None = None,
+        password: Any | None = None,
```
Oleksl888, line 39: shouldn't password be a string?

services/user.py
```diff
@@ -0,0 +1,56 @@
+from typing import Any
+
+from django.contrib.auth import get_user_model
+
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: Any,
```
Oleksl888, line 10: Why Any?

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+
+def update_user(
+        user_id: int,
+        username: str | None = None,
+        password: Any | None = None,
+        email: str | None = None,
+        first_name: str | None = None,
+        last_name: str | None = None
+) -> None:
+    user = get_user_model().objects.get(pk=user_id)
```
Oleksl888, line 44: you can reuse your `get_user` function here

Oleksl888 (changes requested):

db/models.py
```diff
@@ -1,4 +1,11 @@
+from typing import Any
+
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
 from django.db import models
+from django.db.models import UniqueConstraint
+
+import settings
```
MNDonut, line 8: it's better to use
because you may want to have settings folder or your project has multiple settings files

db/models.py
```diff
@@ -50,3 +62,70 @@ class MovieSession(models.Model):
...
+        if not (1 <= self.row <= self.movie_session.cinema_hall.rows):
+            raise ValidationError(
+                {"row": f"row number must be in "
+                    f"available range: (1, rows): (1, "
+                    f"{self.movie_session.cinema_hall.rows})"}
+            )
+
+    def save(self, *args, **kwargs) -> Any:
+        self.full_clean()
+        return super().save(*args, **kwargs)
```
MNDonut, line 127: Are you sure it returns it? It returns None actually(you can hover your mouse on super.save and will see which type it should return

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
MNDonut, line 10: str or None annotation. Fix everywhere

services/movie_session.py
```diff
@@ -42,3 +44,9 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
```
MNDonut, line 46: follow dry and use get_movie_sessions

settings.py
```diff
@@ -25,4 +25,9 @@
 
 INSTALLED_APPS = [
     "db",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
 ]
```
MNDonut, line 30: actually at the top should be default django apps, then yours

MNDonut (changes requested):

services/movie.py
Oleksl888: This haven't been added - you need to add annotation that this parameter can also be None

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # with transaction.atomic():
+    order = Order.objects.create(user=user)
+    if date:
+        order.created_at = parse_datetime(date)
```
Oleksl888, line 21: you can pass date as is - no need for conversion here. Don't forget to clean up imports as well

Oleksl888 (changes requested):

services/order.py
Oleksl888: Lol this actually is a read operation that wouldn't need a rollback in case if something happens :)
But it is actually better to wrap the entire function in transaction

Oleksl888 (approved):
Good job! :clap:

PR #864, opened 2024-09-01
solution

HalynaRohatska:


db/models.py
```diff
@@ -50,3 +55,67 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE
```
Oleksl888, line 64: related name missing here

db/models.py
```diff
@@ -50,3 +55,67 @@ class MovieSession(models.Model):
...
+        return str(self.created_at)
+
+
+class Ticket(models.Model):
+    row = models.IntegerField()
+    seat = models.IntegerField()
+    movie_session = models.ForeignKey(
+        to=MovieSession,
+        on_delete=models.CASCADE,
+        related_name="movie_sessions"
```
Oleksl888, line 80: this is wrong related name

db/models.py
```diff
@@ -50,3 +55,67 @@ class MovieSession(models.Model):
...
+    )
+    order = models.ForeignKey(
+        to=Order,
+        on_delete=models.CASCADE,
+        related_name="orders"
```
Oleksl888, line 85: related name is how `tickets` will be accessed from `Order` model. Please correct

db/models.py
```diff
@@ -50,3 +55,67 @@ class MovieSession(models.Model):
...
+    def __str__(self) -> str:
+        return (f"{self.movie_session.movie.title} "
+                f"{self.movie_session.show_time} "
+                f"(row: {self.row}, seat: {self.seat})")
+
+    def clean(self) -> None:
+        max_rows = self.movie_session.cinema_hall.rows
+        max_seats = self.movie_session.cinema_hall.seats_in_row
+
+        if max_rows < self.row:
```
Oleksl888, line 97: this doesn't validate vs negative numbers

db/models.py
```diff
@@ -50,3 +55,67 @@ class MovieSession(models.Model):
...
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["row", "seat", "movie_session"],
+                name="unique_ticket_seat_row_movie_session")
```
Oleksl888, line 116: well done on naming contraint in a right way :clap: 
But please move Meta class right under fields definition - it is placed there as a convetion in django

services/movie_session.py
```diff
@@ -42,3 +42,9 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    tickets = Ticket.objects.filter(movie_session_id=movie_session_id)
+
+    return [{"row": ticket.row, "seat": ticket.seat} for ticket in tickets]
```
Oleksl888, line 50: you can get result in one line and use `.values()`

services/order.py
```diff
@@ -0,0 +1,45 @@
...
+            order.created_at = date
+            order.save()
+
+        for ticket in tickets:
+            movie_session = MovieSession.objects.get(
+                id=ticket["movie_session"]
+            )
+            Ticket.objects.create(
+                order=order,
+                movie_session=movie_session,
```
Oleksl888, line 30: you can pass movie_session_id instead of passing an object here

services/user.py
```diff
@@ -0,0 +1,63 @@
...
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+
+
+def get_user(user_id: int) -> User:
+    return get_user_model().objects.filter(id=user_id).first()
```
Oleksl888, line 31: use pk here

services/user.py
```diff
@@ -0,0 +1,63 @@
...
+
+def update_user(
+        user_id: int,
+        username: str | None = None,
+        password: str | None = None,
+        email: str | None = None,
+        first_name: str | None = None,
+        last_name: str | None = None
+) -> None:
+    user = get_user_model().objects.filter(id=user_id).first()
```
Oleksl888, line 42: you can reuse your get_user function here

services/user.py
```diff
@@ -0,0 +1,63 @@
...
+
+    if not user:
```
Oleksl888, line 44: no need for this check - if user doesnt exist let the error happen

settings.py
```diff
@@ -24,5 +24,28 @@
...
 
 INSTALLED_APPS = [
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
     "db",
 ]
+
+AUTH_USER_MODEL = "db.User"
+
+AUTH_PASSWORD_VALIDATORS = [
```
Oleksl888, line 34: you don't need to add this in this task

Oleksl888 (changes requested):

db/models.py
```diff
@@ -1,4 +1,9 @@
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
 from django.db import models
+from django.db.models import UniqueConstraint
+
+import settings
```
MNDonut, line 6: it's better to use `from django.conf import settings` because you may have settings folder of another settings file and python will import incorrect settings

db/models.py
```diff
@@ -50,3 +55,68 @@ class MovieSession(models.Model):
...
+    movie_session = models.ForeignKey(
+        to=MovieSession,
+        on_delete=models.CASCADE,
+        related_name="movie_session_tickets"
+    )
+    order = models.ForeignKey(
+        to=Order,
+        on_delete=models.CASCADE,
+        related_name="order_tickets"
+    )
```
MNDonut, line 87: you overwhelm it too much
You have MovieSession object. For example

`my_session = MovieSession.objects.filter(show_date='2024-09-03', movie='Harry Potter').first()`
I know it's MovieSession, now I want to get all tickets of this movie
`my_session.movie_session_tickets.all() # it's a bit redundant to add "movie_session" in related_name because I know that my_session is MovieSession object`
so instead it's better to do
`my_session.tickets.all()`

Fix everywhere ;)

db/models.py
```diff
@@ -50,3 +55,68 @@ class MovieSession(models.Model):
...
+                f"(row: {self.row}, seat: {self.seat})")
+
+    def clean(self) -> None:
+        max_rows = self.movie_session.cinema_hall.rows
+        max_seats = self.movie_session.cinema_hall.seats_in_row
+
+        if max_rows < self.row or self.row < 0:
+            raise ValidationError(
+                {"row": [f"row number must be in available range: "
+                         f"(1, rows): (1, {self.row - 1})"]}
```
MNDonut, line 108: what if row is 0? 0 < 0 will not return false and raise error but you mention here that row should be at least 1

db/models.py
```diff
@@ -50,3 +55,68 @@ class MovieSession(models.Model):
...
+            )
+        if max_seats < self.seat or self.seat < 0:
```
MNDonut, line 110: the same here

services/movie_session.py
MNDonut: still not fixed

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
```
MNDonut, line 13: datetime or None

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+                movie_session_id=ticket["movie_session"],
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+        return order
+
+
+def get_orders(
+        username: str | None = None,
+) -> QuerySet:
```
MNDonut, line 36: specify what kind of data this queryset contains

MNDonut (changes requested):

services/movie.py
```diff
@@ -25,16 +30,17 @@ def get_movie_by_id(movie_id: int) -> Movie:
...
+        movie = Movie.objects.create(
+            title=movie_title,
+            description=movie_description,
+        )
+        if genres_ids:
+            movie.genres.set(genres_ids)
+        if actors_ids:
+            movie.actors.set(actors_ids)
+
+        return movie
```
oliasymak24, line 46: You don't need return anything in create/update/delete functions, check everywhere

oliasymak24 (changes requested):

MNDonut (approved):

PR #865, opened 2024-09-01
Task Solution

achalaidiuk:


db/models.py
```diff
@@ -50,3 +53,75 @@ class MovieSession(models.Model):
...
+    order = models.ForeignKey(
+        to=Order, on_delete=models.CASCADE, related_name="tickets"
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["movie_session", "row", "seat"], name="unique_ticket"
```
Oleksl888, line 86: constraint name should contain names of fields that are unique

db/models.py
```diff
@@ -50,3 +53,75 @@ class MovieSession(models.Model):
...
+        ]
+
+    def __str__(self) -> str:
+        movie = self.movie_session.movie.title
+        movie_session = self.movie_session.show_time
+        return (f"{movie} {movie_session}"
+                f" (row: {self.row}, seat: {self.seat})")
+
+    def clean(self) -> None:
+        super().clean()
```
Oleksl888, line 97: you actually don't need to call super() here because models.Model method clean is just a placeholder to be overriden in child methods

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+            order=order,
+            row=ticket.get("row"),
+            seat=ticket.get("seat")
+        )
+
+
+@transaction.atomic()
+def get_orders(username: str | None = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
```
Oleksl888, line 33: can you refactor it to have only one return instruction?

db/models.py
```diff
@@ -50,3 +53,75 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
```
Oleksl888, line 61: i don't see user-related changes  in ` settings.py`
This is probably because you are on master branch. Shouldn't you work in develop branch instead?

Oleksl888 (changes requested):

MNDonut (approved):

PR #868, opened 2024-09-01
'Solution'

ABaidin:


db/models.py
```diff
@@ -50,3 +59,60 @@ class MovieSession(models.Model):
...
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        to=User, on_delete=models.CASCADE, related_name="orders"
+    )
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+    class Meta:
```
Oleksl888, line 77: Meta class should be defined right after fields definition

db/models.py
```diff
@@ -50,3 +59,60 @@ class MovieSession(models.Model):
...
+            })
+
+    def save(self, *args, **kwargs) -> Callable:
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=["movie_session", "row", "seat"],
+                                    name="unique_ticket"),
```
Oleksl888, line 117: constraint name should contain names of fields that are unique + Meta class should be moved above str method

services/movie_session.py
```diff
@@ -42,3 +44,10 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> List[Dict[str, Any]]:
+    return [
+        {"row": ticket.row, "seat": ticket.seat}
```
Oleksl888, line 51: can you use values() method here instead?

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+                movie_session_id=ticket["movie_session"],
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+            )
+
+
+def get_orders(username: str | None = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
```
Oleksl888, line 35: can you refactor this with one return instruction?

Oleksl888 (changes requested):

MNDonut (approved):

PR #870, opened 2024-09-01
Solution

dpiuro:


db/models.py
```diff
@@ -50,3 +56,61 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        "User",
```
Oleksl888, line 64: you should use AUTH USER MODEL from settings

db/models.py
```diff
@@ -50,3 +56,61 @@ class MovieSession(models.Model):
...
+        on_delete=models.CASCADE,
+        related_name="orders")
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+    class Meta:
+        ordering = ("-created_at",)
```
Oleksl888, line 72: Meta class should be defined right after fields by django convention

db/models.py
```diff
@@ -50,3 +56,61 @@ class MovieSession(models.Model):
...
+    seat = models.IntegerField()
+
+    def __str__(self) -> str:
+        return (
+            f"{self.movie_session.movie.title} {self.movie_session.show_time} "
+            f"(row: {self.row}, seat: {self.seat})"
+        )
+
+    def clean(self) -> None:
+        if self.row > self.movie_session.cinema_hall.rows:
```
Oleksl888, line 94: you don't validate if numbers are below zero here

db/models.py
```diff
@@ -50,3 +56,61 @@ class MovieSession(models.Model):
...
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["movie_session", "row", "seat"],
+                name="unique_ticket_constraint"
```
Oleksl888, line 114: constraint name should contain names of fields that are unique + Meta class should be moved above __str__ method

services/movie_session.py
```diff
@@ -42,3 +44,12 @@ def update_movie_session(
...
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    tickets = Ticket.objects.filter(movie_session_id=movie_session_id)
+    taken_seats = [
+        {"row": ticket.row, "seat": ticket.seat}
+        for ticket in tickets
+    ]
+    return taken_seats
```
Oleksl888, line 55: can you do it using .values() on queryset?

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+
+from db.models import Order, Ticket, MovieSession, User
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> Order:
+    user = User.objects.get(username=username)
```
Oleksl888, line 12: you should use `get_user_model()` to get user and this should also be under atomic decorator

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+        if date:
+            order.created_at = date
+            order.save()
+
+        for ticket_data in tickets:
+            movie_session = MovieSession.objects.get(
+                id=ticket_data["movie_session"]
+            )
+            Ticket.objects.create(
+                movie_session=movie_session,
```
Oleksl888, line 25: no need to create movie_session object - you can pass an id on Ticket creation instead

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+            )
+
+    return order
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        user = User.objects.get(username=username)
+        orders = Order.objects.filter(user=user)
+    else:
```
Oleksl888, line 38: refactor so you use only one `if` statement and return intruction. Use lookup to filter data `user__username`

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+                movie_session=movie_session,
+                order=order,
+                row=ticket_data["row"],
+                seat=ticket_data["seat"]
+            )
+
+    return order
+
+
+def get_orders(username: str = None) -> QuerySet:
```
Oleksl888, line 34: Add annotation for optional parameters

services/user.py
```diff
@@ -0,0 +1,59 @@
+from django.contrib.auth import get_user_model
+
+
+User = get_user_model()
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
```
Oleksl888, line 10: add optional annotation

services/user.py
```diff
@@ -0,0 +1,59 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> User:
+    user = User.objects.get(id=user_id)
```
Oleksl888, line 43: reuse your get_user function here to make it DRY

Oleksl888 (changes requested):

services/movie.py
```diff
@@ -1,11 +1,13 @@
 from django.db.models import QuerySet
+from django.db import transaction
 
 from db.models import Movie
 
 
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
Oleksl888, line 10: Annotation that this parameter can be None is missing

db/models.py
Oleksl888: It is possible to combine these two conditions above one Validation Error such as : `if not 0 < x <= y`

services/movie_session.py
```diff
@@ -42,3 +44,13 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    tickets = (
```
Oleksl888, line 50: no need to store tickets in a variable that you will not reuse. just return result

services/order.py
Oleksl888: yeah, my mistake there is no need for transaction in read operation here :blush:

Oleksl888 (changes requested):

services/movie.py
```diff
@@ -38,3 +40,7 @@ def create_movie(
         movie.actors.set(actors_ids)
 
     return movie
```
oliasymak24, line 42: You don't need return anything in create/update/delete functions, check everywhere

oliasymak24 (changes requested):

db/models.py
```diff
@@ -50,3 +57,73 @@ class MovieSession(models.Model):
...
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        to=MovieSession,
+        on_delete=models.CASCADE,
+        related_name="tickets")
+    order = models.ForeignKey(
+        to=Order,
+        on_delete=models.CASCADE,
+        related_name="tickets")
```
MNDonut, line 85: the closing bracket should be on a separate line

MNDonut (approved):

PR #871, opened 2024-09-02
Solution

olena4reunited:


db/models.py
```diff
@@ -1,8 +1,15 @@
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
 from django.db import models
 
+import settings
```
MNDonut, line 5: it's better to use
`from django.conf import settings` because you may want to have settings folder and python will import it instead of real settings

db/models.py
```diff
@@ -1,8 +1,15 @@
...
+
 
 class Genre(models.Model):
-    name = models.CharField(max_length=255, unique=True)
+    name = models.CharField(
+        max_length=255,
+        unique=True
+    )
```
MNDonut, line 12: no need to move it if it fits in one line in lest than 79 characters

services/movie.py
```diff
@@ -1,11 +1,15 @@
...
 
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
+    genres_ids: Optional[list[int]] = None,
+    actors_ids: Optional[list[int]] = None,
+    title: Optional[str] = None,
```
MNDonut, line 12: let's use | operator everywhere instead of Optional because it's an old approach

MNDonut: Fix everywhere

services/movie_session.py
```diff
@@ -13,7 +17,7 @@ def create_movie_session(
     )
 
 
-def get_movies_sessions(session_date: str = None) -> QuerySet:
+def get_movies_sessions(session_date: Optional[str] = None) -> QuerySet:
```
MNDonut, line 20: Specify what kind of objects this queyset contains

MNDonut (changes requested):

Oleksl888 (approved):
Well done! :clap:

PR #872, opened 2024-09-02
Solution

mihavryliuk:


MNDonut (changes requested):

services/movie.py
```diff
@@ -25,16 +30,17 @@ def get_movie_by_id(movie_id: int) -> Movie:
...
+        movie = Movie.objects.create(
+            title=movie_title,
+            description=movie_description,
+        )
+        if genres_ids:
+            movie.genres.set(genres_ids)
+        if actors_ids:
+            movie.actors.set(actors_ids)
 
     return movie
```
oliasymak24, line 46: You don't need return anything in create/udate/delete functions, check everywhere

oliasymak24 (changes requested):

services/movie.py
mihavryliuk: I did not wright this code it was given

db/models.py
```diff
@@ -50,3 +54,67 @@ class MovieSession(models.Model):
...
+
+    class Meta:
+        ordering = ("-created_at",)
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+
+class Ticket(models.Model):
+    movie_session = ForeignKey(MovieSession, on_delete=models.CASCADE)
```
Oleksl888, line 74: related names are missing  + try to be consistent with imports - in this module you use `models.ForeignKey` but only in this line you use `ForeignKey` which also added an extra import. It is not a mistake but it is better to be consistent

db/models.py
```diff
@@ -50,3 +54,67 @@ class MovieSession(models.Model):
...
+
+    def __str__(self) -> str:
+        return (
+            f"{self.movie_session.movie.title} {self.movie_session.show_time} "
+            f"(row: {self.row}, seat: {self.seat})"
+        )
+
+    def clean(self) -> None:
+        cinema_hall = self.movie_session.cinema_hall
+        if self.row < 1 or self.row > cinema_hall.rows:
```
Oleksl888, line 95: you can build an expression like so which is also valid:
`if not 0 < self.row <= self.movie_session.cinema_hall.rows:`

db/models.py
```diff
@@ -50,3 +54,67 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
```
Oleksl888, line 62: related name?

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime | None = None
```
Oleksl888, line 13: here date is actually in string reprsentation + this block is overindented

services/user.py
```diff
@@ -0,0 +1,47 @@
...
+
+def update_user(
+        user_id: int,
+        username: str | None = None,
+        password: str | None = None,
+        email: str | None = None,
+        first_name: str | None = None,
+        last_name: str | None = None
+) -> None:
+    user = get_user_model().objects.get(pk=user_id)
```
Oleksl888, line 36: reuse you get_user method to avoid repetition

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #874, opened 2024-09-02
Solution

AndriiHamasa:


db/models.py
```diff
@@ -1,4 +1,10 @@
+
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
 from django.db import models
+from django.db.models import UniqueConstraint
+
+import settings
```
because you may want to have settings folder or another settings file

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None,
```
MNDonut, line 10: str or None. Fix it everywhere

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+        ticket_objects = []
+        for ticket in tickets:
+            movie_session = MovieSession.objects.get(
+                pk=ticket["movie_session"]
+            )
+            ticket_objects.append(Ticket(
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+                movie_session=movie_session
```
MNDonut, line 26: let's set movie_session_id here

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+            ))
+
+        Ticket.objects.bulk_create(ticket_objects)
+
+
+def get_orders(username: str = None) -> QuerySet:
```
MNDonut, line 32: specify what kind of data this queryset contains. Fix everywhere

MNDonut (changes requested):

services/movie.py
```diff
@@ -25,16 +30,17 @@ def get_movie_by_id(movie_id: int) -> Movie:
...
+        movie = Movie.objects.create(
+            title=movie_title,
+            description=movie_description,
+        )
+        if genres_ids:
+            movie.genres.set(genres_ids)
+        if actors_ids:
+            movie.actors.set(actors_ids)
+
+        return movie
```
oliasymak24, line 46: You don't need to return anything in create/update/delete functions, check everywhere

oliasymak24 (changes requested):

MNDonut (approved):

PR #875, opened 2024-09-02
Solution

pashawarganov:


db/models.py
```diff
@@ -50,3 +54,67 @@ class MovieSession(models.Model):
...
+            })
+
+        if not (1 <= self.seat <= self.movie_session.cinema_hall.seats_in_row):
+            raise ValidationError({
+                "seat": f"seat number must be in available range: "
+                        f"(1, seats_in_row): "
+                        f"(1, {self.movie_session.cinema_hall.seats_in_row})"
+            })
+
+    def save(self, *args, **kwargs) -> Callable:
```
MNDonut, line 118: It's None actually, save doesn't return anything

services/movie.py
```diff
@@ -6,6 +7,7 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
MNDonut, line 10: str or None

services/movie_session.py
```diff
@@ -42,3 +45,11 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
+    return list(
+        Ticket.objects
+        .filter(movie_session_id=movie_session_id)
```
MNDonut, line 53: Or reuse get_movie_session_by_id and use related_name to fetch tickets of this movies session. But this is ok as well

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket, User, MovieSession
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
```
MNDonut, line 12: datetime or None. Fix evewhere annotation

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+) -> Order:
+    with transaction.atomic():
+        order = Order.objects.create(user=User.objects.get(username=username))
```
MNDonut, line 15: Use get_user_model instead of using User explicitly because the purpose of AUTH_USER_MODEL is to not hardcode main model

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+        if date:
+            order.created_at = date
+            order.save()
+
+        new_tickets = [
+            Ticket(
+                movie_session=MovieSession.objects.get(
+                    pk=ticket["movie_session"]
+                ),
```
MNDonut, line 24: it's better to set movie_session_id cuz you already have id + avoid unneccessary sql query

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+                seat=ticket["seat"]
+            )
+            for ticket in tickets
+        ]
+
+        Ticket.objects.bulk_create(new_tickets)
+        return order
+
+
+def get_orders(username: str = None) -> QuerySet:
```
MNDonut, line 36: str or None

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+                    pk=ticket["movie_session"]
+                ),
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+            for ticket in tickets
+        ]
+
+        Ticket.objects.bulk_create(new_tickets)
```
MNDonut, line 32: I liked this approach :)

services/user.py
```diff
@@ -0,0 +1,65 @@
...
+
+from db.models import User
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
```
MNDonut, line 12: Annotation

services/user.py
```diff
@@ -0,0 +1,65 @@
+from __future__ import annotations
+from django.contrib.auth import get_user_model
```
MNDonut, line 2: But here you use get_user_model, isn't it strange? 👀

services/user.py
```diff
@@ -0,0 +1,65 @@
...
+    return get_user_model().objects.get(pk=user_id)
+
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None,
```
MNDonut, line 42: annotation

MNDonut (changes requested):

services/user.py
pashawarganov: Yes, that's because I had python 3.9. Now I have updated it)

MNDonut (approved):

PR #876, opened 2024-09-03
solution

mnezbrytska:


db/models.py
```diff
@@ -50,3 +57,78 @@ class MovieSession(models.Model):
...
+    )
+
+    def __str__(self) -> str:
+        return (
+            f"{self.movie_session.movie.title} "
+            f"{self.movie_session.show_time} "
+            f"(row: {self.row}, seat: {self.seat})"
+        )
+
+    class Meta:
```
Oleksl888, line 98: meta class should be defined right after field definition

services/movie.py
```diff
@@ -6,14 +7,16 @@
 def get_movies(
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
+    title: str = None
```
Oleksl888, line 10: add optional annotation here and everywhere where it is required

services/movie_session.py
```diff
@@ -42,3 +42,8 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    tickets = Ticket.objects.filter(movie_session_id=movie_session_id)
+    return [{"row": ticket.row, "seat": ticket.seat} for ticket in tickets]
```
Oleksl888, line 49: can you call it on MovieSession with related name and use values() method here?

services/order.py
```diff
@@ -0,0 +1,41 @@
+from datetime import datetime
```
Oleksl888, line 1: add blank lines between different import types

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+        username: str,
+        date: str = None
+) -> Order:
+    user = User.objects.get(username=username)
+
+    if date:
+        created_at = datetime.strptime(date, "%Y-%m-%d %H:%M")
+    else:
+        created_at = datetime.now()
+    order = Order.objects.create(user=user, created_at=created_at)
```
Oleksl888, line 22: Here you should not pass created_at - it should be populated automatically if you add auto_now_add.
Then, if date is being passed you can change it on order object

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+            row=ticket_data["row"],
+            seat=ticket_data["seat"]
+        )
+
+    return order
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
```
Oleksl888, line 40: can you refactor this to have only one return?

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> User:
+    user = User.objects.create_user(
+        username=username,
+        password=password,
+        email=email
```
Oleksl888, line 18: email is optional

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+        user.last_name = last_name
+    user.save()
+    return user
+
+
+def get_user(user_id: int) -> User:
+    try:
+        return User.objects.get(id=user_id)
+    except User.DoesNotExist:
+        raise ObjectDoesNotExist(f"User with id {user_id} does not exist.")
```
Oleksl888, line 32: No need to reraise here - just try to return user.

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
+
+
+def get_user(user_id: int) -> User:
+    try:
+        return User.objects.get(id=user_id)
```
Oleksl888, line 30: use pk here

Oleksl888 (changes requested):

db/models.py
```diff
@@ -60,7 +58,7 @@ def __str__(self) -> str:
 
 
 class Order(models.Model):
-    created_at = models.DateTimeField(default=datetime.datetime.now)
+    created_at = models.DateTimeField(auto_now_add=True)
```
MNDonut, line 61: did you push migration that updates this parameter?

db/models.py
```diff
@@ -1,5 +1,3 @@
-import datetime
-
 from django.contrib.auth.models import AbstractUser
 from django.core.exceptions import ValidationError
```
MNDonut, line 2: cannot highlight the correct code but
instead of
`import settings`
use
because you may have a few settings files or a folder in the same way

services/movie.py
```diff
@@ -1,13 +1,15 @@
...
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
-    title: str = None
+    genres_ids: Optional[list[int]] = None,
+    actors_ids: Optional[list[int]] = None,
+    title: Optional[str] = None
```
MNDonut: it's better to use | operator instead of Optional as a moder way to specify not required parameters

MNDonut: Fix everywhere

MNDonut (changes requested):

oliasymak24 (approved):

PR #877, opened 2024-09-03
task done

Sergi0bbb:


db/models.py
```diff
@@ -50,3 +54,75 @@ class MovieSession(models.Model):
...
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField()
+    user = models.ForeignKey(
+        User,
```
Oleksl888, line 66: User should be taken from AUTH USER MODEL as per best practices

db/models.py
```diff
@@ -50,3 +54,75 @@ class MovieSession(models.Model):
...
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField()
```
Oleksl888, line 64: as per requirements:

> datetime field created_at, the datetime when the order was created, set auto_now_add to fill this field automatically during the creation.

Oleksl888 (changes requested):

services/user.py
```diff
@@ -22,16 +27,16 @@ def create_user(
 
 
 def get_user(user_id: int) -> User:
-    return User.objects.get(id=user_id)
+    return get_user_model().objects.get(id=user_id)
```
MNDonut, line 30: pk instead of id

services/order.py
```diff
@@ -26,8 +30,8 @@ def create_order(
...
 
-def get_orders(username: str = None) -> QuerySet:
+def get_orders(username: str | None = None) -> QuerySet:
+    filters = {}
     if username:
-        user = User.objects.get(username=username)
-        return Order.objects.filter(user=user)
-    return Order.objects.all()
+        filters["user__username"] = username
+    return Order.objects.filter(**filters)
```
MNDonut, line 37: You can implement it in 1 line. Each order has a user field that has related_name for Order model. Thus you can find user with that username and use related_name to get all user's orders 👀

services/user.py
```diff
@@ -1,14 +1,19 @@
...
 
 
 def create_user(
         username: str,
         password: str,
-        email: str = None,
-        first_name: str = None,
+        email: str | None = None,
+        first_name: str | None = None,
         last_name: str = None
```
MNDonut, line 11: fix none annotation

MNDonut (changes requested):

db/models.py
```diff
@@ -50,3 +54,75 @@ class MovieSession(models.Model):
...
+    )
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def save(self, **kwargs) -> None:
+        if not self.created_at:
```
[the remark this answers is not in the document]
Oleksl888: Why do you have it here? It is not required in this task and I saw that in another student's solution..

db/models.py
```diff
@@ -50,3 +54,75 @@ class MovieSession(models.Model):
...
+
+    def __str__(self) -> str:
+        return (f"{self.movie_session}"
+                f" (row: {self.row}, seat: {self.seat})")
+
+    def clean(self) -> None:
+        cinema_hall = self.movie_session.cinema_hall
+        errors = {}
+
+        if self.row > cinema_hall.rows:
```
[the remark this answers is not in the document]
Oleksl888: can you make a one-line conditional expression like so:
`if not 0 < self.row <= self.movie_session.cinema_hall.rows:`

services/movie_session.py
```diff
@@ -41,4 +41,13 @@ def update_movie_session(
...
-    MovieSession.objects.get(id=session_id).delete()
+    MovieSession.objects.get(pk=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return list(
+        MovieSession.objects
+        .filter(pk=movie_session_id)
+        .prefetch_related("tickets")
+        .values(row=F("tickets__row"), seat=F("tickets__seat"))
```
Oleksl888, line 52: or you could do something like this 
`list(get_movie_session_by_id(movie_session_id).tickets.values("row", "seat"))`

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
+
+        return order
+
+
+def get_orders(username: str | None = None) -> QuerySet:
+    return get_user_model().objects.get(username=username).orders.all() \
```
Oleksl888, line 33: 1. remember - we don't use `\` for breaking lines as per code style
2. Here you can try different approach: get order queryset, if username is present - filter queryset using lookup on user field, then return queryset

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done! :handshake:

PR #879, opened 2024-09-03
solving

Poznanskyi:


db/models.py
```diff
@@ -50,3 +60,69 @@ class MovieSession(models.Model):
...
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return self.created_at.strftime("%Y-%m-%d %H:%M:%S")
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        to=MovieSession,
+        on_delete=models.CASCADE
```
Oleksl888, line 87: related names missing here

db/models.py
```diff
@@ -50,3 +60,69 @@ class MovieSession(models.Model):
...
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=["movie_session", "row", "seat"],
+                name="unique_movie_session"
```
Oleksl888, line 100: constraint name should contain all fields that this contraint is applied to

db/models.py
```diff
@@ -50,3 +60,69 @@ class MovieSession(models.Model):
...
+            })
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    def __str__(self) -> str:
+        return (
+            self.movie_session.movie.title + " "
+            + self.movie_session.show_time.strftime("%Y-%m-%d %H:%M:%S")
```
Oleksl888, line 126: can you do it without string concatenation?

services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
+    if MovieSession.objects.filter(id=movie_session_id).exists():
```
Oleksl888, line 48: no need for extra if's and prints, just return result - we don't need to try to catch all errors in this scenario

services/order.py
```diff
@@ -0,0 +1,31 @@
...
+        date: str | None = None
+) -> None:
+    user = get_user_model().objects.get(username=username)
+    new_data = {}
+    if date:
+        new_data["created_at"] = date
+
+    with transaction.atomic():
+        order = Order.objects.create(user=user)
+        Order.objects.filter(id=order.id).update(**new_data)
```
Oleksl888, line 20: or you could set created_at on `order` object and then call save()

services/order.py
```diff
@@ -0,0 +1,31 @@
...
+        for ticket in tickets:
+            ticket["movie_session_id"] = ticket["movie_session"]
+            del ticket["movie_session"]
+            Ticket.objects.create(order=order, **ticket)
+
+
+def get_orders(username: str | None = None) -> QuerySet:
+    kwargs = {}
+    if username:
+        kwargs["user"] = get_user_model().objects.get(username=username)
```
Oleksl888, line 30: can you do this with lookup on Order?

services/user.py
```diff
@@ -0,0 +1,68 @@
...
+
+def create_user(
+        username: str,
+        password: str,
+        email: str | None = None,
+        first_name: str | None = None,
+        last_name: str | None = None
+) -> None:
+    if get_user_model().objects.filter(username=username).exists():
+        print("Username already exists")
```
Oleksl888, line 14: no need for this check

services/user.py
```diff
@@ -0,0 +1,68 @@
...
+
+    get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        **users_data
+    )
+
+
+def get_user(user_id: int) -> User:
+    return get_user_model().objects.filter(pk=user_id).first()
```
Oleksl888, line 33: or you can use .get() here

services/user.py
```diff
@@ -0,0 +1,68 @@
...
+
+def update_user(
+    user_id: int,
+    username: str | None = None,
+    password: str | None = None,
+    email: str | None = None,
+    first_name: str | None = None,
+    last_name: str | None = None
+) -> None:
+    if not get_user_model().objects.filter(pk=user_id).exists():
```
Oleksl888, line 44: no need for this - if user not exists let the exception happen, and later when you work with django you may use get_object_or_404() method for cases like this. For now lets keep it simple

services/user.py
```diff
@@ -0,0 +1,68 @@
...
+        print("User not found")
+        return
+
+    user = get_user_model().objects.get(pk=user_id)
```
Oleksl888, line 48: you can reuse your get_user function here

services/user.py
```diff
@@ -0,0 +1,68 @@
...
+
+    if username:
+        if (
+            user.username != username
+            and get_user_model().objects.filter(username=username).exists()
+        ):
+            print("Username already exists")
```
Oleksl888, line 55: no need for this checks really

Oleksl888 (changes requested):

db/models.py
```diff
@@ -84,7 +84,8 @@ def __str__(self) -> str:
 class Ticket(models.Model):
     movie_session = models.ForeignKey(
         to=MovieSession,
-        on_delete=models.CASCADE
+        on_delete=models.CASCADE,
+        related_name="tickets"
```
MNDonut, line 88: did you push migrations that will update database?

MNDonut (changes requested):

MNDonut (approved):

PR #880, opened 2024-09-03
ready solution!

job-artem:


services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```

services/movie_session.py
```diff
@@ -42,3 +42,10 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list:
+    return list(Ticket.objects
+                .filter(movie_session_id=movie_session_id)
+                .values("row", "seat")
+                )
```
Dimosphen1, line 51: In case there are no arguments on the same line with the closing parenthesis, your opening parenthesis should not have any arguments on the same line as well. Please, resolve this issue

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+
+from db.models import Order, User, Ticket
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
+    with transaction.atomic():
```
Dimosphen1, line 12: The same relates here as well

services/user.py
```diff
@@ -0,0 +1,48 @@
...
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = User.objects.get(pk=user_id)
+    if username is not None:
```
Dimosphen1, line 38: You could simplify similar if conditions in all project
```suggestion
    if username:
```

Dimosphen1 (changes requested):

db/models.py
```diff
@@ -26,6 +29,21 @@ def __str__(self) -> str:
         return self.title
 
 
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
```
pavlopro, line 38: add related name

db/models.py
```diff
@@ -50,3 +68,41 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
```
pavlopro, line 75: Add related names

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+from db.models import Order, User, Ticket
+
+
+@transaction.atomic
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
+) -> None:
+    user = User.objects.get(username=username)
```
pavlopro, line 13: Please use `get_user_model()` function from django to get the auth class from settings

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+    order = Order.objects.create(user=user)
+    if date:
+        parsed_date = parse_datetime(date)
```
pavlopro, line 16: No need to parse it. Simplify to ` order.created_at = date`

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+            row=ticket["row"],
+            seat=ticket["seat"],
+            movie_session_id=ticket["movie_session"]
+        )
+
+
+def get_orders(username: str = None) -> list[Order]:
+    if username is None:
+        return Order.objects.all()
+    user = User.objects.get(username=username)
```
pavlopro, line 32: Simplify to 
```
orders = Order.objects.all()
if username:
return orders
```

services/user.py
```diff
@@ -0,0 +1,48 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> None:
+    user = User.objects.get(pk=user_id)
```
pavlopro, line 37: Use `get_user` method here

pavlopro (changes requested):

PR #881, opened 2024-09-04
implement (first try)

Maksym-Turenko:


services/order.py
```diff
@@ -0,0 +1,40 @@
...
+from db.models import Order, Ticket
+
+
+User = get_user_model()
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str = None
```
MNDonut, line 14: str | none

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+) -> Order:
+    user = User.objects.get(username=username)
+    new_data = {}
+    if date:
+        new_data["created_at"] = date
```
MNDonut, line 19: no need to create a dict for just 1 field

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+
+    with transaction.atomic():
+        order = Order.objects.create(
+            user=user
+        )
```
MNDonut, line 24: you can fit it in 1 line

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+            Ticket.objects.create(
+                order=order,
+                row=ticket_data["row"],
+                seat=ticket_data["seat"],
+                movie_session_id=ticket_data["movie_session"]
+            )
+    return order
+
+
+def get_orders(username: str = None) -> list[Order]:
```
MNDonut, line 37: str | None

MNDonut: Are you sure it returns list?

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+    if username:
+        return Order.objects.filter(user__username=username)
+    return Order.objects.all()
```
MNDonut, line 40: it's better to create a variable and filter it if username is passed

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+
+User = get_user_model()
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = "",
+        last_name: str = ""
```
MNDonut, line 13: str or None

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+        return None
+
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
```
MNDonut, line 38: str or None

settings.py
```diff
@@ -25,4 +25,8 @@
 
 INSTALLED_APPS = [
     "db",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
```
MNDonut, line 29: base apps should at the top of this constant

services/order.py
Maksym-Turenko: It's ain't work into another way 😭

services/movie.py
```diff
@@ -1,13 +1,15 @@
...
 from db.models import Movie
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
-    title: str = None
+    genres_ids: Optional[list[int]] = None,
+    actors_ids: Optional[list[int]] = None,
+    title: Optional[str] = None
```
MNDonut, line 12: let's use a modern way of defining non required parameters using | operator

services/order.py
```diff
@@ -9,19 +11,17 @@
...
-        username: str,
-        date: str = None
+    tickets: list[dict],
+    username: str,
+    date: Optional[str] = None
 ) -> Order:
     user = User.objects.get(username=username)
     new_data = {}
     if date:
         new_data["created_at"] = date
```
MNDonut: isn't it better to create order here and if date -> set date? instead of creating this dict

MNDonut (changes requested):

db/models.py
```diff
@@ -50,3 +53,80 @@ class MovieSession(models.Model):
...
+                }
+            )
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+
+class User(AbstractUser):
+    pass
```
Oleksl888, line 129: if you define __str__ method below you don't need a `pass` keyword here. 
Also `str` method of BaseUser is returning username which is inherited here - so you don't need to overwrite it to do the same thing

services/movie_session.py
```diff
@@ -13,32 +15,47 @@ def create_movie_session(
...
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    taken_seats = Ticket.objects.filter(
+        movie_session_id=movie_session_id
+    ).values("row", "seat")
+
+    return list(taken_seats)
```
Oleksl888, line 61: you can do this in one operation without storing it in variable

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+User = get_user_model()
+
+
+def create_user(
+    username: str,
+    password: str,
+    email: str | None = None,
+    first_name: str = "",
```
Oleksl888, line 12: `None` is preferable default value for optional parameters + here you need to add annotation that it is optional same as above

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+    last_name: str = ""
+) -> User:
+    user = User.objects.create_user(
+        username=username,
+        password=password,
+        email=email,
```
Oleksl888, line 18: you need to write optional values only if they are present - if you fill out your records with garbage values (like with empty strings here) it may lead to incorrect results in future queries. For example you want to filter items where first/last name is not NULL but it will give you incorrect result since you put empty strings there

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+        first_name=first_name,
+        last_name=last_name
+    )
+    return user
+
+
+def get_user(user_id: int) -> User | None:
+    try:
+        return User.objects.get(id=user_id)
```
Oleksl888, line 27: use pk instead

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+    except ObjectDoesNotExist:
+        return None
```
Oleksl888, line 29: if User doesn't exist - let the error happen, it would make sense to stop program execution here instead of dealing with None values

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+            user.set_password(password)
+        if email:
+            user.email = email
+        if first_name:
+            user.first_name = first_name
+        if last_name:
+            user.last_name = last_name
+        user.save()
+        return user
+    except ObjectDoesNotExist:
```
Oleksl888, line 54: same - try except is not necessary. In django you have function get_object_or_404 which is widely used for this and it actually returns an error to signify that something is wrong

Oleksl888 (changes requested):

MNDonut (approved):

PR #883, opened 2024-09-04
Resolve

ihorutkin:


db/models.py
```diff
@@ -42,11 +46,84 @@ def __str__(self) -> str:
...
+                    "seat": [f"seat number must be in available "
+                             f"range: (1, seats_in_row): (1, {seats})"]
+                }
+            )
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
```
oliasymak24, line 123: class Meta should be placed before methods. Don't forget to read checklist, please

services/movie.py
```diff
@@ -28,13 +32,14 @@ def create_movie(
...
+        movie = Movie.objects.create(
+            title=movie_title,
+            description=movie_description,
+        )
+        if genres_ids:
+            movie.genres.set(genres_ids)
+        if actors_ids:
+            movie.actors.set(actors_ids)
+
+        return movie
```
oliasymak24, line 45: You don't need to return anything in create/update/delete functions, check everywhere

services/movie_session.py
```diff
@@ -42,3 +42,17 @@ def update_movie_session(
...
+        movie_session_id=movie_session_id
+    )
+
+    return [
+        {
+            "row": ticket.row,
+            "seat": ticket.seat
+        }
+        for ticket in tickets
+    ]
```
oliasymak24, line 58: You can use .values() here

oliasymak24 (changes requested):

services/movie_session.py
MNDonut: Then you can convert queryset to list 👀

services/movie.py
```diff
@@ -1,9 +1,11 @@
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
+    title: str | None = None,
     genres_ids: list[int] = None,
     actors_ids: list[int] = None,
```
MNDonut, line 10: fix none annotation

services/movie.py
```diff
@@ -1,9 +1,11 @@
...
 ) -> QuerySet:
```
MNDonut, line 11: specify what type of data this queryset contains

services/movie.py
```diff
@@ -28,13 +32,12 @@ def create_movie(
     genres_ids: list = None,
     actors_ids: list = None,
```
MNDonut, line 33: fix annotation

services/movie.py
```diff
@@ -28,13 +32,12 @@ def create_movie(
...
 ) -> Movie:
```
MNDonut, line 34: are you sure this method returns Movie?

services/movie_session.py
MNDonut, line 58: By the way you can do it using 1 ORM call. Fetch movie sesson, then tickets using related name and then .values() method

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+            )
+
+
+def get_orders(username: str | None = None) -> QuerySet:
```
MNDonut, line 35: specify queryset type

MNDonut (changes requested):

MNDonut (approved):

PR #885, opened 2024-09-06
Implement tickets and orders

Morphin20th:


services/order.py
```diff
@@ -0,0 +1,36 @@
...
+            row=ticket.get("row"),
+            seat=ticket.get("seat"),
+            movie_session=movie_session,
+            order=order
+        )
+
+
+def get_orders(
+        username: str = None,
+) -> QuerySet:
```
pavlopro, line 32: indicate QuerySet of which objects you expect

pavlopro (approved):
GJ!

PR #887, opened 2024-09-10
Develop

weuis:


services/movie.py
```diff
@@ -1,11 +1,16 @@
+from __future__ import annotations
```
pavlopro, line 1: Are you sure you need this?

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+from django.contrib.auth import get_user_model
+from django.db import transaction
+from django.db.models import QuerySet
+
+from db.models import Order, Ticket
+
+
+@transaction.atomic()
+def create_order(
+        tickets: list[dict], username: str, date: datetime | None = None
```
pavlopro, line 14: place each parameter on the new line

pavlopro (changes requested):

services/order.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
```
pavlopro, line 1: How about those one?

services/user.py
```diff
@@ -0,0 +1,60 @@
+from __future__ import annotations
```
pavlopro, line 1: How about those one?

db/migrations/0002_user_order_rename.py
```diff
@@ -0,0 +1,100 @@
+# Generated by Django 4.0.2 on 2024-09-10 19:25
```
pavlopro, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid migrations folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #898, opened 2024-09-24
'Solution'

AlexTarasov57:


services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```
pavlopro, line 36: Please apply it as decorator

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+
+from db.models import Ticket, Order
+
+
+def create_order(
+        tickets: list[Ticket],
+        username: str,
+        date: str = None
+) -> Order:
+    with transaction.atomic():
```
pavlopro, line 13: Please apply it as decorator

services/order.py
```diff
@@ -0,0 +1,37 @@
...
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+                movie_session_id=ticket["movie_session"],
+            )
+    return order
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
+    with transaction.atomic():
```
pavlopro, line 32: There no need to have this context manager here

services/user.py
```diff
@@ -0,0 +1,58 @@
...
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> User:
+    user = {
+        "username": username,
+        "password": password,
+    }
+
+    if email is not None:
```
pavlopro, line 16: Simplify it. No need to check those arguments for None.
Use like this:
```
if email:
   ...
```
and so on..

services/user.py
```diff
@@ -0,0 +1,58 @@
...
+        user["email"] = email
+
+    if first_name is not None:
+        user["first_name"] = first_name
+
+    if last_name is not None:
+        user["last_name"] = last_name
+
+    users = User.objects.create_user(**user)
```
pavlopro, line 25: not users, but user

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #903, opened 2024-10-06
Solution

SkepskyiDanylo:


services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
+        movie = Movie.objects.create(
+            title=movie_title,
+            description=movie_description,
+        )
+        if genres_ids:
+            movie.genres.set(genres_ids)
+        if actors_ids:
+            movie.actors.set(actors_ids)
```
vsmutok, line 44: Use transaction.atomic as decorator instead of context manager. It will make your code more readable

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #916, opened 2024-10-20
Solution

Serhii-Chubur:


services/user.py
```diff
@@ -0,0 +1,54 @@
...
+    return get_user_model().objects.get(pk=user_id)
+
+
+def update_user(user_id: int,
+                username: str = None,
+                password: str = None,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None) -> User:
+    user = get_user_model().objects.get(pk=user_id)
```
vsmutok, line 37: you can use `get_user` here

services/order.py
```diff
@@ -0,0 +1,45 @@
...
+                 username: str,
+                 date: str = None) -> None:
+
+    with transaction.atomic():
+
+        user = get_user_model().objects.get(username=username)
+        order = Order.objects.create(user=user)
+        tickets_instances = []
+        if date:
+            order.created_at = date
```
vsmutok, line 16: Use transaction.atomic as decorator instead of context manager. It will make your code more readable

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
+        movie = Movie.objects.create(
+            title=movie_title,
+            description=movie_description,
+        )
+        if genres_ids:
+            movie.genres.set(genres_ids)
+        if actors_ids:
+            movie.actors.set(actors_ids)
+
+        return movie
```
vsmutok, line 46: Use transaction.atomic as decorator instead of context manager. It will make your code more readable

vsmutok (approved):

PR #925, opened 2024-11-03
solution

dmytrik:


db/models.py
```diff
@@ -50,3 +59,59 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = ForeignKey(AUTH_USER_MODEL,
+                      on_delete=models.CASCADE,
+                      related_name="orders")
```
Mxbely: Code style: Use `models.` at the all document or import all fields

db/models.py
```diff
@@ -50,3 +59,59 @@ class MovieSession(models.Model):
...
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
Mxbely: do you need to use f-string here?

db/models.py
```diff
@@ -50,3 +59,59 @@ class MovieSession(models.Model):
...
+
+class Ticket(models.Model):
+    movie_session = ForeignKey(MovieSession,
+                               on_delete=models.CASCADE,
+                               related_name="tickets")
+    order = ForeignKey(Order,
+                       on_delete=models.CASCADE,
+                       related_name="tickets")
+    row = models.IntegerField()
+    seat = models.IntegerField()
```
Mxbely: Same: `models.`

db/models.py
```diff
@@ -50,3 +59,59 @@ class MovieSession(models.Model):
...
+
+    def clean(self) -> None:
+        if not (1 < self.row <= self.movie_session.cinema_hall.rows):
+            raise ValidationError(
+                {"row": ["row number must be in available range: (1, rows): "
+                         f"(1, {self.movie_session.cinema_hall.rows})"]}
+            )
```
Mxbely: bad code style

db/models.py
```diff
@@ -50,3 +59,59 @@ class MovieSession(models.Model):
...
+        if not (1 < self.seat <= self.movie_session.cinema_hall.seats_in_row):
+            raise ValidationError(
+                {"seat": ["seat number must be in available range: "
+                          "(1, seats_in_row): (1, "
+                          f"{self.movie_session.cinema_hall.seats_in_row})"]}
+            )
```
Mxbely: bad code style

db/models.py
```diff
@@ -50,3 +59,59 @@ class MovieSession(models.Model):
...
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        return super().save(*args, **kwargs)
```
Mxbely: return -> None

services/movie_session.py
```diff
@@ -13,7 +13,7 @@ def create_movie_session(
     )
 
 
-def get_movies_sessions(session_date: str = None) -> QuerySet:
+def get_movies_sessions(session_date: str = None) -> QuerySet[MovieSession]:
     queryset = MovieSession.objects.all()
     if session_date:
         queryset = queryset.filter(show_time__date=session_date)
```
Mxbely: where is return?

services/movie_session.py
```diff
@@ -42,3 +42,15 @@ def update_movie_session(
...
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    movie_session = MovieSession.objects.get(pk=movie_session_id)
+
+    return [
+        {
+            "row": ticket.row,
+            "seat": ticket.seat
+        }
+        for ticket in movie_session.tickets.all()
+    ]
```
Mxbely: if you want, you can use `list(... .values("row", "seat"))` for return

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+        first_name: str = None,
+        last_name: str = None
+) -> User:
+    return get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        email=email or "",
+        first_name=first_name or "",
+        last_name=last_name or ""
+    )
```
Mxbely, line 19: if you don't have some fields, you writing `""` in the field, but it is not empty field
if this fields = None, you don't need set it
for example -> see `update_user()`

Mxbely (changes requested):
Not bad, but review some moments

services/movie_session.py
dmytrik: think about it

PR #926, opened 2024-11-04
Solution

VladyslavBon:


services/movie_session.py
```diff
@@ -42,3 +42,10 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict[str, int]]:
+    return [
+        {"row": ticket.row, "seat": ticket.seat}
+        for ticket in Ticket.objects.filter(movie_session_id=movie_session_id)
+    ]
```
Mxbely: You can use `.values` insted of listcomprehation

services/order.py
```diff
@@ -0,0 +1,42 @@
...
+            row=ticket["row"],
+            seat=ticket["seat"],
+            movie_session_id=ticket["movie_session"],
+            order=order
+        )
+        for ticket in tickets
+    ]
+
+    for ticket in tickets:
+        ticket.full_clean()
```
Mxbely, line 34: You can not use `fullclean` for each ticket

Mxbely (approved):

PR #929, opened 2024-11-07
tickets-and-orders solution

94nj111:


services/order.py
```diff
@@ -0,0 +1,38 @@
...
+        date: datetime = None
+) -> Order:
+    with transaction.atomic():
+        order = Order()
+        user = User.objects.get(username=username)
+        order.user = user
+        if date:
+            order.save()
+            order.created_at = date
+        order.save()
```
KKvladys, line 21: i think we dont need use  order.save() 2 time here, but idk

services/order.py
```diff
@@ -0,0 +1,38 @@
...
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> Order:
+    with transaction.atomic():
+        order = Order()
+        user = User.objects.get(username=username)
+        order.user = user
```
KKvladys: we can get user and after create Order with user=user

db/models.py
```diff
@@ -50,3 +53,62 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        to="User",
+        related_name="orders",
```
KKvladys, line 63: the lecture says it's better to access the settings here. (settings.AUTH_USER_MODEL) not to="User"

KKvladys (approved):

PR #930, opened 2024-11-08
Solution

Atikiho:


services/user.py
```diff
@@ -0,0 +1,46 @@
...
+    email: str = None,
+) -> User:
+    user_data = {"username": username, "password": password}
+    if email:
+        user_data["email"] = email
+    if first_name:
+        user_data["first_name"] = first_name
+    if last_name:
+        user_data["last_name"] = last_name
+    return get_user_model().objects.create_user(**user_data)
```
Arsen-hrynevych, line 20: you’re passing password as part of user_data. However, create_user() in Django expects password to be set separately, as it hashes it.

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+    order = Order.objects.create(user=get_user_model().
+                                 objects.get(username=username))
+    if date:
+        order.created_at = date
+    [Ticket.objects.create(
+        movie_session=MovieSession.objects.get(id=ticket["movie_session"]),
+        row=ticket["row"],
+        seat=ticket["seat"],
+        order=order
+    )for ticket in tickets]
```
Arsen-hrynevych, line 26: you’re using a list comprehension to create Ticket instances. List comprehensions are generally meant for generating lists, not for executing functions with side effects.

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None,
+
+) -> None:
+    order = Order.objects.create(user=get_user_model().
+                                 objects.get(username=username))
+    if date:
+        order.created_at = date
```
Arsen-hrynevych, line 20: created_at is typically an auto-populated field in Django

settings.py
```diff
@@ -25,4 +27,6 @@
 
 INSTALLED_APPS = [
     "db",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
```
Arsen-hrynevych, line 31: ...and DB should be the last one

Arsen-hrynevych (approved):
ok

PR #931, opened 2024-11-08
solution

popeye88:


services/order.py
```diff
@@ -0,0 +1,59 @@
...
+            row=row,
+            seat=seat
+        ))
+    return session_tickets
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
+    if username is None:
+        return Order.objects.all().values()
+    else:
```
vsmutok: `else ` redundant in this case

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #934, opened 2024-11-10
Develop

MatvijSafonov:


dmytrik (approved):

db/models.py
```diff
@@ -50,3 +54,65 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        to=settings.AUTH_USER_MODEL,
```
AnyoneClown, line 62: Read documentation. Import `settings` form `django.conf`

[Link](https://docs.djangoproject.com/en/5.1/topics/settings/#using-settings-in-python-code)

services/user.py
```diff
@@ -0,0 +1,59 @@
...
+        password=password
+    )
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+    user.save()
+    return user
```
AnyoneClown, line 26: `create_user` will validate empty fields, so you don't need to check if they are not `None`.

services/user.py
```diff
@@ -0,0 +1,59 @@
...
+        user_id: int,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        email: Optional[str] = None,
+        first_name: Optional[str] = None,
+        last_name: Optional[str] = None
+) -> Optional[User]:
+    try:
+        user = User.objects.get(id=user_id)
+        if username:
```
AnyoneClown, line 46: Use `get_user` method here

settings.py
```diff
@@ -25,4 +25,8 @@
 
 INSTALLED_APPS = [
     "db",
```
AnyoneClown, line 27: `db` should be last in this list

AnyoneClown (changes requested):

PR #940, opened 2024-11-10
solution

spa7id:


services/order.py
```diff
@@ -0,0 +1,39 @@
...
+
+        return order
+
+
+def get_orders(username: str = None) -> Order:
+    if username:
+        user = User.objects.get(username=username)
+        return Order.objects.filter(user=user)
+    else:
+        return Order.objects.all()
```
YaroslavBordovoy: The else block is unnecessary here, it is enough to put a return at the end

services/user.py
```diff
@@ -0,0 +1,50 @@
...
+        user_data["first_name"] = first_name
+    if last_name:
+        user_data["last_name"] = last_name
+
+    some_user = User.objects.create_user(**user_data)
+    return some_user
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
```
YaroslavBordovoy: If a non-existent id is passed, the program will crash. Use get_object_or_404

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+            order.save()
+
+        # Створення квитків
+        for ticket_data in tickets:
+            Ticket.objects.create(
+                order=order,
+                movie_session_id=ticket_data["movie_session"],
+                row=ticket_data["row"],
+                seat=ticket_data["seat"]
+            )
```
YaroslavBordovoy: It may be preferable to use bulk_create() here since we will only make one database query.

YaroslavBordovoy (changes requested):
In new models, specify related_name, it won't be superfluous.
Remove the changes in the tests, perhaps you accidentally made them.

PR #946, opened 2024-11-11
Add new migrations

yulia-pl:


services/movie_session.py
```diff
@@ -1,44 +1,37 @@
...
+    session_id: int, **kwargs
+) -> MovieSession:
     movie_session = MovieSession.objects.get(id=session_id)
-    if show_time:
-        movie_session.show_time = show_time
-    if movie_id:
-        movie_session.movie_id = movie_id
-    if cinema_hall_id:
-        movie_session.cinema_hall_id = cinema_hall_id
+    for key, value in kwargs.items():
```
dxrrkwm, line 22: didn't really need to change it according to readme but it's a better way to use loop here i guess :)

dxrrkwm (approved):
looks good i think😇

Arsen-hrynevych (approved):
ok

PR #986, opened 2024-12-17
Solution

Icemann007:


db/migrations/0001_initial.py
vsmutok, line 1: Please do not modify the existing migration file. It's better to create a new migration file that contains the new changes and leave the old one untouched.

vsmutok (changes requested):

Icemann007:
I can't restore the original file with 1 migration because I closed the project, so I'm pushing 2 migrations. I understand that this is wrong, but I can't fix it, without 1 migration, the tests don't work

vasylhnatiuk (approved):

PR #1001, opened 2025-01-09
Solution

bknyrik:


services/user.py
```diff
@@ -0,0 +1,58 @@
...
+def update_user(
+    user_id: int,
+    username: str | None = None,
+    password: str | None = None,
+    email: str | None = None,
+    first_name: str | None = None,
+    last_name: str | None = None
+) -> None:
+
+    user = User.objects.get(pk=user_id)
```
vsmutok: better to use `get_user_model()` here

services/user.py
```diff
@@ -0,0 +1,58 @@
...
+    if first_name:
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+
+
+def get_user(user_id: int) -> User:
```
vsmutok, line 28: better to use `get_user_model()` here

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1012, opened 2025-01-14
Solution

AntonBessarab:


services/user.py
```diff
@@ -0,0 +1,49 @@
...
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+
+
+def get_user(user_id: int) -> AbstractBaseUser:
```
vsmutok: The `get` method will raise a `DoesNotExist` exception if the user is not found. Consider handling this exception to avoid runtime errors.

services/user.py
```diff
@@ -0,0 +1,49 @@
...
+
+def update_user(
+        user_id: int,
+        username: str = None,
+        password: str = None,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None,
+) -> None:
+    user = get_user_model().objects.get(id=user_id)
```
vsmutok: use `get_user` here

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
+        movie = Movie.objects.create(
+            title=movie_title,
+            description=movie_description,
+        )
+        if genres_ids:
+            movie.genres.set(genres_ids)
+        if actors_ids:
+            movie.actors.set(actors_ids)
+
+        return movie
```
vsmutok: use transaction.atomic decorator for better readability

vsmutok (changes requested):

SerhiyTo (approved):

PR #1014, opened 2025-01-15
Solution

IvankaKuzin:


services/movie.py
```diff
@@ -15,10 +17,15 @@ def get_movies(
     if actors_ids:
         queryset = queryset.filter(actors__id__in=actors_ids)
 
+    if title:
+        queryset = queryset.filter(title__icontains=title)
+
     return queryset
```
Dogober, line 23: where do you filter movies by genres_ids

services/movie_session.py
```diff
@@ -40,5 +46,23 @@ def update_movie_session(
...
+) -> list[dict]:
+    result = []
+
+    movie_session = MovieSession.objects.get(id=movie_session_id)
+
+    for tickets in movie_session.tickets.all():
+        result.append({
+            "row": tickets.row,
+            "seat": tickets.seat,
+        })
```
Dogober: You can use list comprehension here

services/order.py
```diff
@@ -0,0 +1,48 @@
...
+
+        new_order.save()
+
+        for ticket in tickets:
+            new_ticket = Ticket.objects.create(
+                movie_session=MovieSession.objects.get(
+                    id=ticket["movie_session"]
+                ),
+                order_id=new_order.id,
+                row=ticket["row"],
```
Dogober: Use the @transaction.atomic decorator instead of using with transaction.atomic: for better readability and to ensure the entire function executes within a single transaction context
This information from checklist. Read 6 point

services/order.py
```diff
@@ -0,0 +1,48 @@
...
+
+def get_orders(
+        username: str = None
+) -> QuerySet:
+    if username:
+        orders = Order.objects.filter(user__username=username)
+    else:
+        orders = Order.objects.all()
+
+    return orders
```
Dogober: You can simplify this func -> return Order.objects.all() instead of adding aditional else block

Dogober (approved):

services/movie.py
IvankaKuzin: In 15th line :)

PR #1018, opened 2025-01-15
Solution

bodiakof:


services/movie.py
```diff
@@ -1,21 +1,15 @@
...
-    queryset = Movie.objects.all()
-
-    if genres_ids:
-        queryset = queryset.filter(genres__id__in=genres_ids)
-
-    if actors_ids:
-        queryset = queryset.filter(actors__id__in=actors_ids)
-
-    return queryset
+def get_movies(title: Optional[str] = None) -> QuerySet[Movie]:
```
4ikkamony: It was not necessary to remove filtering by genres_ids and actors_ids

services/movie.py
4ikkamony, line 1: Should have made create_movie a transaction, 
It was at the end of the task)

4ikkamony (approved):
Nice work!
Just fix the flake errors and good to go

bodiakof:
> Nice work! Just fix the flake errors and good to go

Will do 💪 Thanks for the review!

PR #1019, opened 2025-01-15
Add user, order, and ticket models with related services

Bleit1:
This commit introduces custom `User`, `Order`, and `Ticket` models, along with updates to the Movie model to include a `db_index` on the `title` field. It also adds related services for managing users, orders, tickets, and their interactions. Necessary settings and migrations are included to support the new models and constraints.


10kkyvl (approved):

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+def create_order(
+        tickets: dict,
+        username: str,
+        date: str = None
+) -> None:
+    user = get_user_model().objects.get(username=username)
+
+    order = Order.objects.create(user=user)
+    if date:
+        _date = datetime.datetime.strptime(date, "%Y-%m-%d %H:%M")
```
L1nk3rrr, line 19: do you really need another one variable? You can overwrite your `date` as it's not used anymore.

L1nk3rrr (approved):
Well done!

PR #1025, opened 2025-01-15
Solution

kuinorymes:


bogdAAAn1 (approved):
Good job!

db/models.py
```diff
@@ -50,3 +59,69 @@ class MovieSession(models.Model):
...
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["movie_session", "row", "seat"],
+                name="unique_ticket"
+            )
+        ]
+
```
petrykivd: class Meta should be placed after all attributes and before methods.

services/movie.py
```diff
@@ -6,13 +7,14 @@
...
-
-    if genres_ids:
+    if title is not None:
+        queryset = queryset.filter(title__icontains=title)
+    if genres_ids is not None:
         queryset = queryset.filter(genres__id__in=genres_ids)
-
-    if actors_ids:
+    if actors_ids is not None:
         queryset = queryset.filter(actors__id__in=actors_ids)
```
petrykivd: if title is the same as if title is not None 😄

services/movie_session.py
```diff
@@ -41,4 +43,13 @@ def update_movie_session(
 
 
 def delete_movie_session_by_id(session_id: int) -> None:
-    MovieSession.objects.get(id=session_id).delete()
+    try:
+        movie_session_to_delete = MovieSession.objects.get(id=session_id)
+        movie_session_to_delete.delete()
+    except MovieSession.DoesNotExist:
+        print(f"No Movie session with id {session_id}")
```
petrykivd: You can avoid additional movie_session_to_delete variable, MovieSession.objects.get(id=session_id).delete()

services/order.py
```diff
@@ -0,0 +1,47 @@
...
+                new_order.save()
+
+            for ticket in tickets:
+                try:
+                    movie_session = MovieSession.objects.get(
+                        id=ticket["movie_session"]
+                    )
+                except MovieSession.DoesNotExist:
+                    print(f"No movie session with id: {ticket['id']} found")
+                    continue
```
petrykivd: Are you sure that in the context of a transaction, creating an order should skip sessions that do not exist? Then this order will not be complete.
In this case, it will be normal if the transaction stops and the order is not created.

services/order.py
```diff
@@ -0,0 +1,47 @@
...
+                Ticket.objects.create(
+                    order=new_order,
+                    movie_session=movie_session,
+                    row=ticket["row"],
+                    seat=ticket["seat"]
+                )
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
+    if username is not None:
```
petrykivd: Same here about is not None

petrykivd: Fix in other places too

petrykivd (changes requested):

petrykivd (approved):
Good job! 👍

PR #1026, opened 2025-01-15
Solution_tickets-and-orders

SummerKatrin:


db/models.py
```diff
@@ -50,3 +52,59 @@ class MovieSession(models.Model):
...
+                f"(row: {self.row}, seat: {self.seat})")
+
+    def clean(self) -> None:
+        if self.row > self.movie_session.cinema_hall.rows:
+            raise ValidationError({
+                "row": f"row number must be in available range: "
+                       f"(1, rows): (1, {self.movie_session.cinema_hall.rows})"
+            })
+
+        if self.seat > self.movie_session.cinema_hall.seats_in_row:
```
MaksVakulenko, line 94: there you need to validation that self.seat < 1

db/models.py
```diff
@@ -50,3 +52,59 @@ class MovieSession(models.Model):
...
+    row = models.IntegerField()
+    seat = models.IntegerField()
+
+    def __str__(self) -> str:
+        return (f"{self.movie_session.movie.title} "
+                f"{self.movie_session.show_time} "
+                f"(row: {self.row}, seat: {self.seat})")
+
+    def clean(self) -> None:
+        if self.row > self.movie_session.cinema_hall.rows:
```
MaksVakulenko, line 88: or self.row < 1

db/models.py
```diff
@@ -50,3 +52,59 @@ class MovieSession(models.Model):
...
+                        f"(1, seats_in_row): "
+                        f"(1, {self.movie_session.cinema_hall.seats_in_row})"
+            })
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    class Meta:
+        unique_together = ("movie_session", "row", "seat")
```
MaksVakulenko, line 106: you need to use UniqueConstraint but not unique_together (read documentation)

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+        first_name = ""
+
+    if last_name is None:
+        last_name = ""
+
+    User.objects.create_user(username=username,
+                             password=password,
+                             email=email,
+                             first_name=first_name,
+                             last_name=last_name)
```
MaksVakulenko, line 21: there just you need first at all create user and after check if first name(etc) is not none. else use -> user.first_name = first_name (etc)

MaksVakulenko (approved):
check my coments

L1nk3rrr (changes requested):
Agreed with comments from AI and Maks, review all of them. 
Take a close look at comments that I liked

PR #1028, opened 2025-01-16
solution

MaksVakulenko:


db/models.py
```diff
@@ -50,3 +60,61 @@ class MovieSession(models.Model):
...
+
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(fields=["row", "seat"], name="unique_ticket")
+        ]
+
```
petrykivd: class Meta shoukd be placed after attributes and before methods.

petrykivd (changes requested):

petrykivd (approved):
Nice! 😎

PR #1029, opened 2025-01-17
solution

yelizaveta-zh:


db/models.py
```diff
@@ -50,3 +53,70 @@ class MovieSession(models.Model):
...
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+
+    def __str__(self) -> str:
+        return self.created_at.strftime("%Y-%m-%d %H:%M:%S")
+
+    class Meta:
+        ordering = ["-created_at"]
+
```
petrykivd, line 68: class Meta should be placed after attributes and before methods. Don't forget to read checklist, please
Fix it everywhere

db/models.py
```diff
@@ -50,3 +53,70 @@ class MovieSession(models.Model):
...
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        to=MovieSession, on_delete=models.CASCADE, related_name="tickets"
+    )
+    order = models.ForeignKey(
+        Order,
+        on_delete=models.CASCADE,
+        related_name="tickets"
+    )
```
petrykivd: Fix codestyle

services/movie_session.py
```diff
@@ -42,3 +40,14 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(
+        movie_session_id: MovieSession
```
petrykivd: movie_session_id is MovieSession? Or it is int? 😄

services/user.py
```diff
@@ -0,0 +1,50 @@
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
```
petrykivd: We don't capitalize common variables. 
User must be constant ( USER ) or user.

petrykivd (changes requested):

petrykivd (approved):
Nice! 😄

PR #1030, opened 2025-01-17
wrong solution

shretsker:


4ikkamony (changes requested):
Seems like you've got a problem with this task

if you have questions, feel wree to write in QnA chat or in telegram, I (and others, I'm sure, too) will be glad to help out

Good luck!

4ikkamony (approved):
Nice work!

There are still a few minor things(like __str__ should return type str, not None, and some other type hint issues)

Apart from that - all good

PR #1078, opened 2025-02-25
Solution

Baranotik15:


db/models.py
```diff
@@ -50,3 +57,58 @@ class MovieSession(models.Model):
...
+
+class User(AbstractUser):
+    pass
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
+    row = models.IntegerField()
+    seat = models.IntegerField()
```
Boyok1337: I recommend using `PositiveIntegerField` here, read about it in the documentation.

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+        if date:
+            order.created_at = date
+            order.save(update_fields=["created_at"])
+
+        for ticket_data in tickets:
+            Ticket.objects.create(
+                row=ticket_data["row"],
+                seat=ticket_data["seat"],
+                movie_session_id=ticket_data["movie_session"],
+                order=order
```
Boyok1337: What if `ticket_data` have no data and you got `KeyError`? Replace with `.get()`

Boyok1337 (changes requested):

db/models.py
```diff
@@ -50,3 +57,58 @@ class MovieSession(models.Model):
...
+
+class User(AbstractUser):
+    pass
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession, on_delete=models.CASCADE)
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
+    row = models.PositiveIntegerField()
+    seat = models.PositiveIntegerField()
```
Boyok1337, line 84: GJ! But do not forget to make a new migration with this changes and push, it important!

Boyok1337 (approved):

PR #1085, opened 2025-02-28
Solution

kram3ko:
flake8 ignore settings.py cause of long lines.


.flake8
```diff
@@ -9,3 +9,4 @@ exclude =
     venv
     .venv
     tests
+    settings.py
```
Boyok1337: Bad practice, I understand that you didn't change the settings, but you should try to format and practice formatting

db/models.py
```diff
@@ -50,3 +75,50 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(
+        to=MovieSession, on_delete=models.CASCADE
+    )
```
Boyok1337: each arg from new line

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+        order.created_at = date
+        order.save()
+
+    for ticket in tickets:
+        Ticket.objects.create(
+            row=ticket["row"],
+            seat=ticket["seat"],
+            movie_session_id=ticket["movie_session"],
+            order=order,
+        )
```
Boyok1337: What if `ticket` have no data and you got KeyError? Replace with .get()

services/user.py
```diff
@@ -0,0 +1,50 @@
...
+@transaction.atomic
+def create_user(
+    username: str,
+    password: str,
+    email: str = None,
+    first_name: str = "",
+    last_name: str = "",
+) -> None:
+    user_model = get_user_model()
+    user_model.objects.create_user(
```
Boyok1337: Better to move it from func to const value on global level of this file. Like before this func like this `User=get_user_model()` instead of `user_model = get_user_model()`
And reuse in each func, instead of call in every next func

services/user.py
```diff
@@ -0,0 +1,50 @@
...
+        password=password,
+        email=email,
+        first_name=first_name,
+        last_name=last_name,
+    )
+
+
+def get_user(user_id: int) -> AbstractBaseUser:
+    user_model = get_user_model()
+    return user_model.objects.get(id=user_id)
```
Boyok1337: same here

Boyok1337 (changes requested):

Y-Havryliv (approved):

PR #1094, opened 2025-03-05
solution

yarzhuravlov:


services/order.py
```diff
@@ -0,0 +1,50 @@
...
+    movie_session = MovieSession.objects.get(pk=tickets[0]["movie_session"])
+
+    tickets = Ticket.objects.bulk_create(
+        [
+            Ticket(
+                row=ticket_dict["row"],
+                seat=ticket_dict["seat"],
+                movie_session=movie_session,
+                order=order,
+            )
```
kram3ko, line 34: What if ticket have no data and you got KeyError? you can replace it  for  .get()

kram3ko (approved):
In general Goodjob

services/order.py
yarzhuravlov: If the data in the ticket dictionary is invalid, we should throw an error, because if we don't, data consistency will be broken. Throwing an error when the dictionary is invalid, combined with the transaction mechanism, ensures that the order will only be created if valid ticket data has been passed.

PR #1095, opened 2025-03-11
Solution

ip-ilya:


services/movie_session.py
```diff
@@ -31,6 +31,7 @@ def update_movie_session(
     cinema_hall_id: int = None,
 ) -> None:
     movie_session = MovieSession.objects.get(id=session_id)
```
SerhiyTo: I strongly recommend using `get_object_or_404() ` instead of the usual `get()`. This will help prevent a situation where the execution of this method may fail without handling an error.

SerhiyTo (changes requested):

PR #1100, opened 2025-03-16
Solution

MarianaAB19:


services/order.py
```diff
@@ -0,0 +1,27 @@
+from django.contrib.auth import get_user_model
+from django.db import transaction
+from django.db.models import QuerySet
+from db.models import Order, Ticket, User
+
+
+@transaction.atomic
+def create_order(tickets: list, username: str, date: str = None) -> list:
+    user = User.objects.get(username=username)
```
vsmutok: It's better to use` get_user_model()` instead of `User`, it is the best practice.

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1105, opened 2025-03-21
Completed task py-tickets-and-orders

Yurkiv313:


db/models.py
```diff
@@ -50,3 +58,71 @@ class MovieSession(models.Model):
...
+        on_delete=models.CASCADE,
+
+    )
+    order = models.ForeignKey(
+        Order,
+        on_delete=models.CASCADE,
+        related_name="tickets"
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
```
Boyok1337: Probably PositiveIntegerField is better here

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+from db.models import Ticket, Order, MovieSession
+
+
+@transaction.atomic
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> Order:
+    user = get_user_model().objects.get(username=username)
```
Boyok1337: Define `User = get_user_model() ` on global level and reuse it

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+        ticket_objects.append(ticket_)
+
+    Ticket.objects.bulk_create(ticket_objects)
+
+    return order
+
+
+def get_orders(username: str = None) -> Order:
+    if username:
+        user = get_user_model().objects.get(username=username)
```
Boyok1337: *to the previous comment

services/user.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None
+) -> User:
+    user = get_user_model().objects.create_user(
```
Boyok1337: The same

services/user.py
```diff
@@ -0,0 +1,54 @@
...
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+    return user
+
+
+def get_user(user_id: int) -> User:
+    return get_user_model().objects.get(pk=user_id)
```
Boyok1337: The same

Boyok1337 (changes requested):

services/order.py
```diff
@@ -0,0 +1,43 @@
...
+) -> Order:
+    user = User.objects.get(username=username)
+    order = Order.objects.create(user=user)
+    if date:
+        order.created_at = date
+        order.save()
+
+    ticket_objects = []
+    for ticket in tickets:
+        movie_session = MovieSession.objects.get(id=ticket["movie_session"])
```
oliasymak24: Use movie_session_id argument and pass the id directly.

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1116, opened 2025-03-27
Solution

OlenVe:


services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
 ) -> Movie:
-    movie = Movie.objects.create(
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
+    with transaction.atomic():
```
AndyStarGitHub, line 36: In the requirements they asked to use @transaction. atomic, not context manager.

AndyStarGitHub (approved):
Good job! I just added one comment for future.

services/movie.py
OlenVe: Thank you for comment. I'll fix it

OlenVe:
@mateacademy-ai-mentor
"services/user.py:
In the create_user function, you're calling user.save() multiple times. It's best practice to call it only once after you've set all the attributes. This can improve performance and prevent unintended side effects."

Its not True. I called save() only once.

PR #1123, opened 2025-03-27
Solution

Kart0l:


PrimeGlorious (changes requested):
You have to push **settings.py** with installed apps for User, i can't verify the request

PR #1128, opened 2025-03-29
Anton Zhukov - Solution

windings-lab:


windings-lab:
@mateacademy-ai-mentor 
Because of auto_now_add=True
I need to call save twice. How do I fix it?

PR #1136, opened 2025-03-31
'Solution'

pewlay:


db/models.py
```diff
@@ -1,4 +1,8 @@
+import settings
```
Boyok1337: Do not use settings directly, use it via django

db/models.py
```diff
@@ -42,11 +51,74 @@ def __str__(self) -> str:
...
+        MovieSession,
+        on_delete=models.CASCADE,
+        related_name="sessions"
+    )
+    order = models.ForeignKey(
+        Order,
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+    row = models.IntegerField()
```
Boyok1337: You can also use PositiveIntegerField or PositiveSmallIntegerField

db/models.py
```diff
@@ -42,11 +51,74 @@ def __str__(self) -> str:
...
+    seat = models.IntegerField()
```
Boyok1337: The same

Boyok1337 (changes requested):

Boyok1337 (approved):
LGTM!

PR #1151, opened 2025-04-13
Implemented Tickets and Orders models, and services for them

Tolik-code:


farzamgt (approved):
Well done

Tolik-code:
@mateacademy-ai-mentor The Genre model doesn't have an f-string in my code. And The string representation for the Ticket also doesnt have angle bruckets, check again

PR #1226, opened 2025-06-20
Develop

alxitk:


services/user.py
```diff
@@ -0,0 +1,54 @@
+from db.models import User
+
+
+def create_user(username: str,
+                password: str,
+                email: str = None,
+                first_name: str = None,
+                last_name: str = None
+                ) -> User:
```
Dimosphen1: In case there are no arguments on the same line with a closing parenthesis, your opening parenthesis should not have any arguments on the same line as well. Please, resolve this issue here and in the other places as well, e.g. on line 27 in `services/order.py`

services/order.py
```diff
@@ -0,0 +1,35 @@
+from datetime import datetime
+from typing import List
+
+from django.db import transaction
+from django.db.models import QuerySet
+from db.models import Ticket, Order, User, MovieSession
+
+
+def create_order(tickets: List[dict], username: str, date: str = None) -> None:
+    with transaction.atomic():
```
Dimosphen1: In case you apply atomic transaction to the whole function body, you can use it as a decorator instead of context manager

Dimosphen1 (changes requested):
Several minor changes were requested

vsmutok (approved):
Looks Good!

PR #1228, opened 2025-06-22
add task solution

EugeneSherepa:


services/movie.py
```diff
@@ -1,9 +1,13 @@
+from typing import Optional
+
+from django.db import transaction
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
+    title: Optional[str] = None,
```
MNDonut: use | operator instead everywhere

services/movie_session.py
```diff
@@ -42,3 +42,17 @@ def update_movie_session(
...
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    movie_session = MovieSession.objects.get(id=movie_session_id)
+    tickets = Ticket.objects.filter(movie_session=movie_session)
+
+    tickets = [
+        {
+            "row": ticket.row,
+            "seat": ticket.seat
+        }
+        for ticket in tickets]
```
MNDonut: use `values_list` ORM method instead

services/order.py
```diff
@@ -0,0 +1,34 @@
...
+            order.created_at = date
+            order.save()
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                movie_session_id=ticket["movie_session"],
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"]
+            )
```
MNDonut: Recommend you to use `.bulk_create()` method instead because each .create call runs 1 sql query, so if you create 1000 tickets, it'll run 1000 sql queries that is too long
On the other side, `bulk_create` runs only 1 sql query

MNDonut (changes requested):

PR #1246, opened 2025-07-16
Refactor order creation to allow custom created_at date; fix __str__ …

kravetst:
…methods and update tests accordingly


kravetb (approved):
Don't use GBT to much!

PR #1259, opened 2025-08-03
Solution

farzamgt:


db/models.py
```diff
@@ -50,3 +54,73 @@ class MovieSession(models.Model):
...
+                    f"(1, rows): (1, {max_row})"
+                ]
+            })
+        if self.seat > max_seat:
+            raise ValidationError({
+                "seat": [
+                    f"seat number must be in available range: "
+                    f"(1, seats_in_row): (1, {max_seat})"
+                ]
+            })
```
Orixara, line 110: there we can add all cases in dict and if dict is not empry raise ValidationError with this dict, like that:

image

services/user.py
```diff
@@ -0,0 +1,56 @@
...
+    if username:
+        user.username = username
+    if password:
+        user.set_password(password)
+    if email:
+        user.email = email
+    if first_name:
+        user.first_name = first_name
+    if last_name:
+        user.last_name = last_name
```
Orixara, line 53: there we have repeated informations it will be better to make additional function to check if something is None:
image

Orixara (approved):
great job! I sent you some advice to make your code better

PR #1261, opened 2025-08-03
solution py tickets and orders

PavelSemenikhin:


SerhiiBasok (approved):
Pavel, I like your style of code execution. Keep it up!

PR #1262, opened 2025-08-03
Solution

dimashevchukk:


db/models.py
```diff
@@ -22,6 +27,11 @@ class Movie(models.Model):
     actors = models.ManyToManyField(to=Actor, related_name="movies")
     genres = models.ManyToManyField(to=Genre, related_name="movies")
```
reinvvay, line 27: bad practice to use the same related_name for different models

reinvvay (approved):

PR #1263, opened 2025-08-03
add new models and services

SerhiiBasok:


services/order.py
```diff
@@ -0,0 +1,32 @@
...
+                seat=ticket["seat"],
+                movie_session=movie_sessions
+            )
+
+
+def get_orders(username: str = None) -> QuerySet:
+    if username:
+        return Order.objects.filter(user__username=username)
+    else:
+        return Order.objects.all()
```
yagginov:
```suggestion
    if username:
        return Order.objects.filter(user__username=username)
    return Order.objects.all()
```

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+        tickets: list[dict],
+        username: str,
+        date: int = None
+) -> None:
+    with transaction.atomic():
+        user = get_user_model().objects.get(username=username)
+        new_order = Order.objects.create(user=user)
+        if date:
+            new_order.created_at = date
+            new_order.save()
```
yagginov, line 17: So if you don't pass the date to the function, the order won't be saved or what? Move the save one level down, like this:
```suggestion
        if date:
            new_order.created_at = date
        new_order.save()
```

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+        for ticket in tickets:
+            movie_sessions = MovieSession.objects.get(id=1)
```
yagginov: The id for the movie session is taken from the tickets, not just the first one. Now it turns out you can only buy tickets for the first session XD

db/models.py
```diff
@@ -50,3 +59,73 @@ class MovieSession(models.Model):
...
+            f"(row: {self.row}, seat: {self.seat})"
+        )
+
+    def clean(self) -> None:
+        free_seats = self.movie_session.cinema_hall.seats_in_row
+        free_rows = self.movie_session.cinema_hall.rows
+
+        if free_rows < self.row:
+            raise ValidationError(
+                {
```
yagginov: there is no check for trying to buy a zero ticket, or a negative one, for example. If you used models.PositiveIntegerField() then there would be no problems, but there users can buy tickets with negative numbers

services/order.py
```diff
@@ -0,0 +1,32 @@
...
+            Ticket.objects.create(
+                order=new_order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+                movie_session=movie_sessions
+            )
```
yagginov: So if you don't pass the date to the function, the order won't be saved or what?

db/models.py
```diff
@@ -50,3 +59,73 @@ class MovieSession(models.Model):
...
+                    "row": [f"row number must be in available range:"
+                            f" (1, rows):"
+                            f" (1, {free_rows})"]
+                }
+            )
+        if free_seats < self.seat:
+            raise ValidationError(
+                {
+                    "seat": [f"seat number must be in available range:"
+                             f" (1, seats_in_row):"
```
yagginov: similar situation

yagginov (changes requested):
There are some mistakes that are better to fix before AI sees them.

yagginov (approved):

PR #1266, opened 2025-08-03
Solution

yagginov:


MakovM (approved):

yagginov:
@mateacademy-ai-mentor In this task we work only with ORM, so I created a user because I needed to create it for the task, the task did not set any additional requirements regarding which migration it should be in.

PR #1268, opened 2025-08-04
Solution

Gleg0:


VadymKyrylko (approved):
Good work! But what about using str method in 29 line in models.py?

PR #1271, opened 2025-08-04
Solution

Artcrafterrra:


db/models.py
```diff
@@ -50,3 +54,67 @@ class MovieSession(models.Model):
...
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class User(AbstractUser):
+    pass
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(to=User, on_delete=models.CASCADE)
```
dimashevchukk: giving related name is good practice

db/models.py
```diff
@@ -50,3 +54,67 @@ class MovieSession(models.Model):
...
+        on_delete=models.CASCADE,
+        related_name="tickets"
+    )
+    order = models.ForeignKey(
+        Order,
+        on_delete=models.CASCADE,
+        related_name="tickets"
+    )
+    row = models.IntegerField()
+    seat = models.IntegerField()
```
dimashevchukk, line 90: PositiveIntegerField would be better

db/models.py
```diff
@@ -50,3 +54,67 @@ class MovieSession(models.Model):
...
+    movie_session = models.ForeignKey(
+        MovieSession,
+        on_delete=models.CASCADE,
+        related_name="tickets"
+    )
+    order = models.ForeignKey(
+        Order,
+        on_delete=models.CASCADE,
+        related_name="tickets"
+    )
```
dimashevchukk, line 88: using same related name is bad practice

services/movie.py
```diff
@@ -28,13 +33,14 @@ def create_movie(
...
-        title=movie_title,
-        description=movie_description,
-    )
-    if genres_ids:
-        movie.genres.set(genres_ids)
-    if actors_ids:
-        movie.actors.set(actors_ids)
-
-    return movie
+    with transaction.atomic():
```
dimashevchukk, line 36: checklist says about using @transaction.atomic decorator instead of with transaction.atomic

services/movie_session.py
```diff
@@ -42,3 +42,8 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    tick = get_movie_session_by_id(movie_session_id).tickets.all()
```
dimashevchukk, line 48: tick?

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+            order.created_at = date
+            order.save()
+
+        for ticket in tickets:
+            Ticket.objects.create(
+                order=order,
+                movie_session_id=ticket.get("movie_session"),
+                row=ticket.get("row"),
+                seat=ticket.get("seat"),
+            )
```
dimashevchukk, line 26: use bulk_create to not make sql query every iteration

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = "",
+        first_name: str = "",
+        last_name: str = "",
+) -> User:
+    return User.objects.create_user(
```
dimashevchukk, line 11: use get_user_model instead of hardcoded User

dimashevchukk (changes requested):

dimashevchukk (approved):

PR #1275, opened 2025-08-04
Solution with error :(

Arch0998:


db/models.py
```diff
@@ -19,6 +23,7 @@ def __str__(self) -> str:
 class Movie(models.Model):
     title = models.CharField(max_length=255)
```
Plovuezz, line 24: cant see indexing here

Plovuezz (approved):
Overall very good, but pls fix indexing🥺

(In `db/models.py` you already have tables you created earlier. Now
you have to create or edit tables:
1. `Movie`, add:
    - index on field `title`.)

Arch0998:
@mateacademy-ai-mentor You're wrong. If I use auto_now_add=True, the current date will be used, which won't match the test settings, and the tests will fail. Everything is correct on my end. Please accept the pull request.

PR #1280, opened 2025-08-05
tickets and orders

Salmon-eat:


services/user.py
```diff
@@ -0,0 +1,43 @@
...
+        first_name: str = "",
+        last_name: str = ""
+) -> User:
+    return User.objects.create_user(
+        username=username,
+        password=password,
+        email=email,
+        first_name=first_name or "",
+        last_name=last_name or ""
+    )
```
SensibleN00B: Use get_user_model() instead of User, it is the best practice.

Good example:
get_user_model().objects.create_user(
    username=username, password=password
)

Bad example:
User.objects.create_user(
    username=username, password=password
)

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+
+User = get_user_model()
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime = None
+) -> Order:
+    with transaction.atomic():
```
SensibleN00B: Use the @transaction.atomic decorator instead of using with transaction.atomic: for better readability and to ensure the entire function executes within a single transaction context

services/order.py
```diff
@@ -0,0 +1,36 @@
...
+                order=order,
+                row=ticket["row"],
+                seat=ticket["seat"],
+                movie_session_id=ticket["movie_session"]
+            )
+        return order
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
+    with transaction.atomic():
```
SensibleN00B: the same, decorator would be better

SensibleN00B (changes requested):

PR #1281, opened 2025-08-05
Solution

SensibleN00B:


services/order.py
```diff
@@ -0,0 +1,35 @@
...
+        username: str,
+        date: str = None
+) -> None:
+    user = User.objects.get(username=username)
+    order = Order.objects.create(user=user)
+    if date:
+        order.created_at = date
+    else:
+        order.created_at = "2020-11-10 14:40"
+    order.save()
```
Kostenikov: Just remove this hardcoded part and all be good😉
```suggestion
    if date:
        order.created_at = date
        order.save()
```

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+            order=order
+        )
+
+
+def get_orders(username: str = None) -> QuerySet[Order]:
+    if username:
+        user = User.objects.get(username=username)
+        return Order.objects.filter(user=user).select_related("user")
+
+    return Order.objects.select_related("user")
```
Kostenikov: All orders can easy filtering like this:
```suggestion
    orders = Order.objects.all()

    if username:
        orders = orders.filter(user__username=username)

    return orders
```

services/order.py
```diff
@@ -0,0 +1,35 @@
...
+from db.models import Order, User, Ticket
+
+
+@transaction.atomic
+def create_order(
+        tickets: list,
+        username: str,
+        date: str = None
+) -> None:
+    user = User.objects.get(username=username)
```
Kostenikov: You need to use get_user_model() here with `import from django.contrib.auth import get_user_model`
```suggestion
    user = get_user_model().objects.get(username=username)
```

Kostenikov (changes requested):

services/user.py
```diff
@@ -0,0 +1,60 @@
...
+
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+    return user
+
+
+def get_user(user_id: int) -> User:
+    return User.objects.get(id=user_id)
```
Kostenikov: Same here, you need to use get_user_model()
```suggestion
    return get_user_model().objects.get(pk=user_id)
```

Kostenikov (approved):
Well Done!

PR #1293, opened 2025-08-14
Solution

Artstrik:


db/migrations/0003_user_order_alter_movie_title_ticket_and_more.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR in the future, please

services/order.py
```diff
@@ -0,0 +1,65 @@
...
+
+User = get_user_model()
+
+
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: str | None = None,
+) -> Order:
+    with transaction.atomic():
```
Dimosphen1, line 17: In case all body of a method should be treated as an atomic transaction, it is better to use `transaction.atomic` as a decorator

Dimosphen1 (approved):
Looks good to me, well done! Several comments were left for consideration in the future

PR #1333, opened 2025-10-03
Solution

Konstantin-Svt:


Konstantin-Svt:
@mateacademy-ai-mentor cannot change services/order.py: create_order to single Order save() because auto_now_add=True flag in created_at field overrides manual data during the first save.

Konstantin-Svt:
@mateacademy-ai-mentor I will not change create_order function because single save will lose provided datetime data due to auto_now_add=True flag in DateTimeField and fail one of Unittest

PR #1335, opened 2025-10-07
Solution

fabio-Silva-E:


db/models.py
```diff
@@ -50,3 +59,65 @@ class MovieSession(models.Model):
...
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        to=User, on_delete=models.CASCADE, related_name="orders"
+    )
+
+    def __str__(self) -> str:
+        return f"{self.created_at.strftime("%Y-%m-%d %H:%M:%S")}"
```
meruyme, line 71: Just a tip: there's no need to use a f-string in here, since the strftime method already returns a string.

db/models.py
```diff
@@ -8,6 +10,13 @@ def __str__(self) -> str:
         return self.name
 
 
+class User(AbstractUser):
+    bio = models.TextField(blank=True, null=True)
```
meruyme, line 14: There's no need to add a bio field in user model. I noticed that this field is not being used in any part of the code.

meruyme (approved):
Your project is correct, congratulations! I'm approving it, but I recommend that you check the comments left by me and by our AI Mentor, so you may improve your code even more.

db/models.py
meruyme: Also, I just noticed that there's a flake8 complaint in the pipeline of your code. To solve it, just remove the f-string as I mentioned earlier, or do this:

```suggestion
        return f"{self.created_at.strftime('%Y-%m-%d %H:%M:%S')}"
```

PR #1384, opened 2025-11-30
Tickets-and-orders

n-korobko:


settings.py
```diff
@@ -25,4 +25,8 @@
 
 INSTALLED_APPS = [
     "db",
+    "django.contrib.auth",
```
stanislav-kishechko, line 28: Require import this section before own app it`s important for Django application

stanislav-kishechko (approved):

PR #1394, opened 2025-12-02
Develop

Ann-Riabykina:


db/models.py
```diff
@@ -50,3 +61,58 @@ class MovieSession(models.Model):
 
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(User,
```
Vorongor: Here should be a reference to settings.AUTH_USER_MODEL not to just user model

db/models.py
```diff
@@ -50,3 +61,58 @@ class MovieSession(models.Model):
...
+
+class Ticket(models.Model):
+    movie_session = models.ForeignKey(MovieSession,
+                                      on_delete=models.CASCADE,
+                                      related_name="tickets")
+    order = models.ForeignKey(Order,
+                              on_delete=models.CASCADE,
+                              related_name="tickets")
+    row = models.IntegerField()
+    seat = models.IntegerField()
```
Vorongor: Good option here may use PositiveIntegerField, which allows avoiding negative values

services/order.py
```diff
@@ -0,0 +1,33 @@
...
+from db.models import Order, Ticket, User
+
+
+@transaction.atomic
+def create_order(
+        tickets: list,
+        username: str,
+        date: Optional[str] = None
+) -> Order:
+    user = User.objects.get(username=username)
```
Vorongor, line 18: The task means to use a built-in user and method, so you should use here: get_user_model()

services/user.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: Optional[str] = None,
+        first_name: Optional[str] = None,
+        last_name: Optional[str] = None
+) -> User:
+    user = User.objects.create_user(username=username, password=password)
```
Vorongor: Also  using the built-in user model involves - using get_user_model()

Vorongor (approved):
Overall, good job! You avoid using built-in user model, but the rest is good!))

PR #1400, opened 2025-12-03
Solution

tigersaymon:


Akcolotle (approved):

tigersaymon:
I’ve reviewed all your comments and they’re not really valid or relevant, because what’s written in the README, REQUIREMENTS, and CHECKLIST differs from what the PYTEST tests actually check.

I focused on making the program work and pass all the PYTEST tests, so from my side everything is correct and there’s nothing more to fix.

Thanks for understanding.

db/models.py
```diff
@@ -50,3 +53,70 @@ class MovieSession(models.Model):
...
     def __str__(self) -> str:
         return f"{self.movie.title} {str(self.show_time)}"
+
+
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey("User", on_delete=CASCADE, related_name="orders")
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
```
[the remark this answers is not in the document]
tigersaymon: The task required the exact format YYYY-MM-DD HH:MM:SS, without 'Order'. If I use your format, the tests will fail.

db/models.py
```diff
@@ -50,3 +53,70 @@ class MovieSession(models.Model):
...
+    def save(self, *args, **kwargs) -> None:
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    def __str__(self) -> str:
+        return (
+            f"{self.movie_session.movie.title} "
+            f"{self.movie_session.show_time} "
+            f"(row: {self.row}, seat: {self.seat})"
+        )
```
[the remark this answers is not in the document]
tigersaymon: The task required my format, without 'Ticket'. If I use your format, the tests will fail.

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+    user = get_user_model().objects.get(username=username)
+
+    order = Order.objects.create(
+        user=user
+    )
+
+    if date:
+        order.created_at = date
+
+    order.save()
```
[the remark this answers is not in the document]
tigersaymon: No, it's not, in this situation we need another save if the user provide date, because date can be changed

db/models.py
```diff
@@ -50,3 +53,70 @@ class MovieSession(models.Model):
...
+class Order(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey("User", on_delete=CASCADE, related_name="orders")
+
+    def __str__(self) -> str:
+        return f"{self.created_at}"
+
+    class Meta:
+        ordering = ["-created_at"]
+
```
petrykivd: class Meta should be placed before methods, please fix it everywhere

db/models.py
petrykivd, line 118: You can use multiline string here : 
f"""
some text
...
"""

services/movie.py
```diff
@@ -1,14 +1,21 @@
...
 
 
 def get_movies(
-    genres_ids: list[int] = None,
-    actors_ids: list[int] = None,
-) -> QuerySet:
+        title: Optional[str] = None,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None,
+) -> QuerySet[Movie]:
```
petrykivd: only title is optional? Why genres ids and actors ids by default also None ?:)

services/movie.py
```diff
@@ -1,14 +1,21 @@
...
     queryset = Movie.objects.all()
 
+    if title is not None:
+        queryset = queryset.filter(title__icontains=title)
+
```
petrykivd: Can be simplified as i"f title"?

services/movie.py
```diff
@@ -22,11 +29,12 @@ def get_movie_by_id(movie_id: int) -> Movie:
...
+@transaction.atomic
 def create_movie(
-    movie_title: str,
-    movie_description: str,
-    genres_ids: list = None,
-    actors_ids: list = None,
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list = None,
+        actors_ids: list = None,
```
petrykivd: Check your indentation settings

services/order.py
```diff
@@ -0,0 +1,40 @@
...
+def create_order(
+        tickets: list,
+        username: str,
+        date: str = None
+) -> Order:
+    user = get_user_model().objects.get(username=username)
+
+    order = Order.objects.create(
+        user=user
+    )
```
petrykivd: Let's have 1 code style.

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+        first_name: Optional[str] = None,
+        last_name: Optional[str] = None
+) -> User:
+    return get_user_model().objects.create_user(
+        username=username,
+        password=password,
+        email=email if email is not None else "",
+        first_name=first_name if first_name is not None else "",
+        last_name=last_name if last_name is not None else ""
+    )
```
petrykivd, line 21: Simplify checking

services/user.py
```diff
@@ -0,0 +1,55 @@
...
+
+    if first_name:
+        user.first_name = first_name
+
+    if last_name:
+        user.last_name = last_name
+
+    user.save()
+
+    return user
```
petrykivd, line 54: In "create user" func you check all information in another way, let's use one of? :)

petrykivd (changes requested):

db/models.py
tigersaymon: Thanks, but using triple quotes causes the tests to fail due to extra whitespace. I had to keep it as a single line to match the expected output (tests)

petrykivd (approved):
Nice! 🚀

PR #1401, opened 2025-12-03
Solution

Akcolotle:


ohorodnykostap (approved):
The code looks correct but fix the code style according to flake8.

PR #1453, opened 2026-02-12
solution

Olena-Shulga:


Olena-Shulga:
@mateacademy-ai-mentor 
My __str__ methods are implemented correctly, the angle-bracket wrapper and the Order prefix will appear in the finale result because that's the way that django outputs objects. Also, I cannot change the number of save() methods in create_order() method, because the create_at field doesn't change if I don't create the Order object first.

PR #1455, opened 2026-02-14
Code write

vova-Sh2:


vova-Sh2:
@mateacademy-ai-mentor 
For the created_at field with auto_now_add=True, it is impossible to set a value upon creation, so the second call to .save() is necessary for the tests to work correctly and does not violate the DRY principle.

PR #1491, opened 2026-03-31
Solution

AndrewGalynin:


AndrewGalynin:
I tried adding the <Order: ...> format as requested, but it caused the test_order_str and test_ticket_str automated tests to fail. The tests expect a strict string match without any prefixes or brackets. I have reverted to the version that passes all tests

vsmutok (approved):
Good Job!

PR #1493, opened 2026-04-03
Solution

antonb285:


db/migrations/0005_alter_order_user_alter_ticket_movie_session_and_more.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR in the future, please

Dimosphen1 (approved):
Looks good to me, well done! One comment is left for consideration

PR #1517, opened 2026-04-23
task tickets-and-orders done

Oh-no-dotcom:


Boyok1337 (changes requested):
Please, make sure that the code passes all tests before submitting it for review

services/order.py
```diff
@@ -0,0 +1,39 @@
...
+        )
+
+
+def get_orders(
+        username: str = None
+) -> QuerySet[Order]:
+    if username:
+        return Order.objects.filter(user__username=username)
+    else:
+        return Order.objects.all()
```
vsmutok:
```suggestion
        return Order.objects.filter(user__username=username)
    return Order.objects.all()
```

vsmutok (changes requested):

db/models.py
```diff
@@ -50,3 +56,65 @@ class MovieSession(models.Model):
...
+    class Meta:
+        ordering = ["-created_at"]
+
+    def __str__(self) -> str:
+        return str(self.created_at)
+
+
+class Ticket(models.Model):
+    row = models.IntegerField()
+    seat = models.IntegerField()
```
Boyok1337, line 78: You can use PositiveIntegerField too

Boyok1337 (approved):
`get_user_model() `is better define one time within file

PR #1519, opened 2026-04-23
I left 2 comments in db.models & services.order.py cause tests is wrong

pawunder1992:


AnyoneClown (changes requested):
Visin QnA session

AnyoneClown (approved):
LGTM

PR #1532, opened 2026-05-04
implement tickets and orders task

shpejik2:


db/models.py
```diff
@@ -50,3 +57,79 @@ class MovieSession(models.Model):
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="orders"
+    )
+
+    def __str__(self) -> str:
+        time = datetime.strftime(self.created_at, "%Y-%m-%d %H:%M:%S")
+        return f"<Order: {time}>"
```
viktoria-rybenchuk:
```suggestion
        return f"{self.created_at}"
```

viktoria-rybenchuk (changes requested):

services/user.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def create_user(
+        username: str,
+        password: str,
+        email: str = None,
+        first_name: str = None,
+        last_name: str = None,
+) -> User:
+    user = get_user_model().objects.create_user(
```
vsmutok: Do not use `get_user_model()` multiple times.
Declare a variable that stores `get_user_model()` and use it everywhere it is needed.

vsmutok (changes requested):

vsmutok (approved):
Looks Good!

PR #1542, opened 2026-05-07
Solution

herman-py:


Prus-Oleksandr (approved):

services/movie_session.py
```diff
@@ -42,3 +42,11 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict]:
+    return list(
+        Ticket.objects.filter(
+            movie_session_id=movie_session_id
+        ).values("row", "seat")
```
VadimPusiak788: You can use `values_list()` instead of `values()`, and you don't need to wrap the result in `list()`

VadimPusiak788 (changes requested):

VadimPusiak788 (approved):

PR #1563, opened 2026-05-16
Implemented solution for py-tickets-and-orders

Neyrony:


Neyrony:
@mateacademy-ai-mentor i beg you not to check the migrations files it's redundant and useless

PR #1572, opened 2026-05-24
tickets and orders

Shongi:


Shongi:
@mateacademy-ai-mentor 
I used your exact recommendation and the tests fail as you can see...
"created_at_kwargs = {'created_at': datetime.strptime(date, "%Y-%m-%d %H:%M")} if date else {} and then user.orders.create(**created_at_kwargs)"

PR #1582, opened 2026-06-02
Solution

Alex-x000:


services/order.py
```diff
@@ -0,0 +1,29 @@
...
+
+    for data in tickets:
+        Ticket.objects.create(movie_session_id=data["movie_session"],
+                              order=order,
+                              row=data["row"],
+                              seat=data["seat"])
+    return order
+
+
+def get_orders(username: str = None) -> Order:
```
vsmutok, line 26: Return type should be `QuerySet[Order]` not just `Order`.

vsmutok (approved):

PR #1597, opened 2026-06-14
Solution

realasdfg:


realasdfg:
@mateacademy-ai-mentor I can't call the save() method only one time in create_order service function because created_at has auto_now_add=True

PR #1603, opened 2026-06-22
a solution

PEMOHTOKOH:


vsmutok (changes requested):
Tests failed.
Please make sure that your code passes all tests before submitting the task for review.

services/order.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@transaction.atomic
+def create_order(
+    tickets: list[dict[str, int]],
+    username: str,
+    date: datetime = None,
+) -> None:
+
+    user = User.objects.get(username=username)
```
vsmutok, line 16: Using `User.objects.get(username=username)` directly contradicts the requirement to rely on `get_user_model()` (checklist item [#4](https://github.com/mate-academy/py-tickets-and-orders/pull/4) and Previous Review Corrections); switch to `get_user_model().objects.get(...)` or call your `get_user` service function instead.

vsmutok (changes requested):

PR #1612, opened 2026-07-04
Solution

herasymchukkdenysss:


herasymchukkdenysss:
@mateacademy-ai-mentor it is not possible to create order in one save operation if date is provided because when auto_now_add is true django ignore you create_at=value in first object save(creation of an object) so I need first to save row objects and then check on date and save again if it is provided https://docs.djangoproject.com/en/6.0/ref/models/fields/?utm_source=chatgpt.com look it up.

PR #1626, opened 2026-07-23
'Solution'

dem1hh:


services/order.py
```diff
@@ -0,0 +1,38 @@
...
+
+User = get_user_model()
+
+
+@transaction.atomic
+def create_order(
+        tickets: list[dict],
+        username: str,
+        date: datetime.datetime | None = None,
+) -> Any:
```
Sn1kls, line 17: return type is `-> Any` – you could narrow it to `-> Order` for precision

Sn1kls (approved):
Approve, but check comment)

PR #1635, opened 2026-08-06
Develop

askawa:


services/movie_session.py
```diff
@@ -42,3 +46,10 @@ def update_movie_session(
 
 def delete_movie_session_by_id(session_id: int) -> None:
     MovieSession.objects.get(id=session_id).delete()
+
+
+def get_taken_seats(movie_session_id: int) -> list[dict[str, int]]:
+    return list(
+        MovieSession.objects.get(id=movie_session_id)
+        .tickets.values("row", "seat")
+    )
```
Sn1kls, line 55: this does two queries – one to fetch the session, one for its tickets. You can get the same result in a single query straight from `Ticket`:

```
from db.models import Ticket
...
...
return list(
    Ticket.objects.filter(movie_session_id=movie_session_id)
    .values("row", "seat")
)
```

Not a blocker – your version is correct, just a small efficiency note for the future

Sn1kls (approved):
Approve. Nice work!)

